PR #1, opened 2019-04-17
Set up project using operator-sdk

grdryn:
This is really just boiler-y plate stuff I think.

Questions in this PR:
* Is the name of the project satisfactory?
* Is the name of the CRD (`UnifiedPushServer`) satisfactory?

Questions for the next part (can be this PR or a future one):
* Integreatly use Travis for their operator CI jobs, should I add something similar here?
* They also have a Makefile, but I'm not sure of the value of one of those here _yet_, if the only thing it's going to do is wrap `operator-sdk` or similar commands. Thoughts?


wei-lee:
@grdryn the names of the repo and CRD are fine to me.

> Integreatly use Travis for their operator CI jobs, should I add something similar here?

Yes, you can use either travis or Circle CI depending on which one you prefer.

> They also have a Makefile, but I'm not sure of the value of one of those here yet, if the only thing it's going to do is wrap operator-sdk or similar commands. Thoughts?

If you look at the Makefile in keycloak operator, it can be used for variable tasks, like setup the cluster, cleanup the cluster, run tests and build/publish images. These tasks are commonly used and have them in makefile will save time I think. If we don't use it, how do you propose where we add these tasks?

PR #6, opened 2019-04-24
Provision UPS based on CR

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9009

~This doesn't yet create the Ingress or Route, since it's not 100%
clear what we need to do around that, and the interaction with the
oauth-proxy container in the UPS Pod.~

For now this just assumes OpenShift and creates a Route. This gives us something to iterate on, where we can have logic to check if OpenShift or plain K8s and act accordingly.

~Kubernetes doesn't appear to provide an OAuth server ootb like
OpenShift does, so the OpenShift oauth-proxy that we use may not work
there as easily.~

~Also, when setting the OAuthRedirectReference in the ServiceAccount,
the OpenShift docs currently claim that Route is the only supported
Kind at the moment.~

We might be able to use `oauth-redirecturi` for Ingress rather than `oauth-redirectreference`.

~There's also the more obvious differences between Route and Ingress:
we're used to using Route, where we just give it a name and the
fully-qualified URL gets provided to us automatically, and configured
in the Router. For k8s Ingress, we need to provide the host directly
in the Ingress resource, so for the operator to do that, it would need
to know what that should be: so it would probably need to be provided
by the person provisioning, in the UnifiedPushServer CR.~

### Verification steps
1. Checkout this branch
2. As a cluster admin user in OpenShift (because of role creation), run `make cluster/prepare`. This will create a `unifiedpush` project and other things needed for the operator.
3. Deploy the image built for this PR:

```
cat deploy/operator.yaml | sed -e "s|quay.io/aerogear/unifiedpush-operator|quay.io/aerogear/unifiedpush-operator:$(git rev-parse --short HEAD)|g" | kubectl create -f -
```

4. Check that the operator has started successfully with like `kubectl get po` or whatever, and have a look at the logs to make sure there's no obvious errors in there.

5. Create a UPS!

```
kubectl apply -f deploy/crds/aerogear_v1alpha1_unifiedpushserver_cr.yaml
```

6. All of the required resources should get created. ~It might happen that the UPS container itself doesn't get to a ready state, if the PostgreSQL wasn't ready in time for it. Unfortunately for now this means that you'll just have to delete the pod for it (I'll create follow-on issues to both wait for PostgreSQL, and also to add a liveness check to the UPS pod/container). Once it's deleted, a new one will be automatically created.~ I've added a liveness check here now in this PR, so it should get restarted until it works.

7. Once everything is ready, you should be able to login through the oauth-proxy in your browser:

```
xdg-open https://$(oc get route -o json | jq -r .items[0].spec.host)
```


pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -100,51 +103,184 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+										},
+									},
+								},
+								InitialDelaySeconds: 15,
+								TimeoutSeconds:      2,
+							},
+						},
+						{
+							Name:            "ups-oauth-proxy",
+							Image:           "docker.io/openshift/oauth-proxy:v1.1.0", // TODO: allow overriding this
```
grdryn: Should this also be configurable through the CRD like the UPS one? Or is that the right place for defining images at all?

If they should be in the CRD, should it be like:

```yaml 
# [...]
spec:
  upsImage: docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final
  oauthProxyImage: docker.io/openshift/oauth-proxy:v1.1.0
# [...]
```
or more like:
```yaml 
# [...]
spec:
  images:
    unifiedpush: docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final
    oauthproxy: docker.io/openshift/oauth-proxy:v1.1.0
# [...]
```

or something else?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -100,51 +103,184 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	// Define a new Pod object
-	pod := newPodForCR(instance)
+	secret := newPostgresSecret(instance)
```
grdryn: This stuff all needs to be cleaned up, it was just getting something working (mostly)

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -100,51 +103,184 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		err = r.client.Create(context.TODO(), secret)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Define a new Deployment object
+	deployment := newDeploymentForCR(instance)
```
grdryn: I'm guessing it's fine that an operator creates a deployment and watches that. It could alternatively create the pods (or ReplicaSets) directly and manage those itself (like how ReplicaSet (or Deployment) controllers already do. I'm not sure which is better?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
wei-lee: I like the second option. I think these values should be used as default if none is provided. Use the values from the CR file if they are defined. This will allow us to override them to use the productised images if required.

wei-lee: Actually, think more about it, I think these values should be set via env vars, rather than in the CR. The reason is that these values are not something that can be changed by the CRs: what happens when there are 2 CRs specify different values for the images? One of them surely will not work. But if we use env var, it won't be a problem.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
wei-lee: One of the difference between Deployment and ReplicaSet is to Deployment can control how rolling updates are performed. So if you use Deployment, you can configure it there, otherwise you will need to implement the updates in your code too?

grdryn: Yeah, it seems natural to use Deployment, but I guess I was just doubting myself, given that the operator-sdk default creates a Pod. I'm sure there are some other cases where actually doing it all in the operator gives people more flexibility or something, but we probably don't need that

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Yeah, that's a good idea I think. Where would we specify those env vars though? In the Dockerfile for the operator build?

Edit: in that case they'd also need to be set for the `make code/run` target somehow, since that runs the operator locally rather than building the image: the env vars would also need to be available for that case.

grdryn:
@wei-lee if you haven't already, could you please also take a look at the general questions in the PR description here. Thanks!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
wei-lee: I think it is ok to set the default values in the code, and only checks the env var when it is set. If nothing is set just use the default one. This way we don't need to define them in dockerfiles, but can still override them in other environments.

wei-lee:
@grdryn 

> Is representing all of the k8s objects like this in code the best
way, or would it be better to create yaml files, and just load those
from the filesystem?

It is ok to write them in the code, and it is my preference at least, because you can get code completion, compared to writing pure yaml files. Writing in yaml is hard to test as you won't know if it's working until you run it.

> Should the operator create the secret at all, or just require the
user to have created that beforehand? If the operator should create
the secret, is there a recommended way for generating passwords that
we should use?

It is ok for operators to generate secrets, and users shouldn't need to create them before hand. Some services allow users to specify passwords in the CR (encoded some how), but the downside is that other users may be able to see it. So it depends on the use cases I think. To generate those passwords, they are not used for encryption or anything so just generate some random strings is enough for this purpose.

grdryn:
Thanks for all the feedback Wei! :+1:

grdryn:
> > Should the operator create the secret at all, or just require the
> > user to have created that beforehand? If the operator should create
> > the secret, is there a recommended way for generating passwords that
> > we should use?
> 
> It is ok for operators to generate secrets, and users shouldn't need to create them before hand. Some services allow users to specify passwords in the CR (encoded some how), but the downside is that other users may be able to see it. So it depends on the use cases I think. To generate those passwords, they are not used for encryption or anything so just generate some random strings is enough for this purpose.

For this case at least, if other users are able to see it in the CR, then they'd also be able to see it in the Secret, so I don't think that's a huge issue. I don't think it really matters what the password is, so it'll be better if we just generate one I think. I had a look at the keycloak operator, and I'll just copy the password generation function from there for now: https://github.com/integr8ly/keycloak-operator/blob/master/pkg/keycloak/keycloak.go#L169..L175

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -100,53 +102,162 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
 
-		// Pod created successfully - don't requeue
-		return reconcile.Result{}, nil
+	postgresqlDeployment, err := newPostgresqlDeployment(instance)
```
grdryn, line 164: From here to L150 is basically a copypasta of the above logic, and from L150 down it's mostly just repeating the same-ish logic over and over for each resource that's being created. I spent a while trying to create a generic function that could do it for all objects, but I was unsuccessful (my Go-fu is not strong enough).

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -52,9 +48,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &aerogearv1alpha1.UnifiedPushServer{},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to secondary resource Secrets and requeue the owner UnifiedPushServer
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForOwner{
```
grdryn: Argh, need more watch-y watch-y.

pkg/controller/unifiedpushserver/util.go
```diff
@@ -0,0 +1,55 @@
...
+// objectMeta returns the default ObjectMeta for all the other objects here
+func objectMeta(cr *aerogearv1alpha1.UnifiedPushServer, suffix string) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      fmt.Sprintf("%s-%s", cr.Name, suffix),
+		Namespace: cr.Namespace,
+		Labels:    labels(cr, suffix),
+	}
+}
+
+func generatePassword() (string, error) {
```
grdryn, line 49: I just plagiarized this function from the keycloak-operator.

grdryn (commented):
I'm also not really sure the files that I've split this into is a good way of doing it. It could just as easily be more files or fewer files. :duck:

grdryn:
Here's more discussion about the Ingress/Route/oauth-proxy stuff that I mentioned in the (updated) description: https://issues.jboss.org/browse/AEROGEAR-8987?focusedCommentId=13727199

grdryn:
@secondsun @danielpassos @wei-lee this much is in a reviewable/runnable state now, if you'd like to take a look. As mentioned in the description, I'll create some follow-up issues for things that should be improved.

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -0,0 +1,207 @@
...
+								},
+							},
+							Args: []string{
+								"--provider=openshift",
+								fmt.Sprintf("--openshift-service-account=%s", cr.Name),
+								"--upstream=http://localhost:8080",
+								"--http-address=0.0.0.0:4180",
+								"--skip-auth-regex=/rest/sender,/rest/registry/device,/rest/prometheus/metrics,/rest/auth/config",
+								"--https-address=",
+								"--cookie-secret=SECRET",
```
grdryn: ~The value here should be generated, rather than the hardcoded "SECRET" from the example. The [docs](https://github.com/openshift/oauth-proxy#configuration) suggest using something equivalent to `python -c 'import os,base64; print base64.b64encode(os.urandom(16))'`.~ Done

pkg/controller/unifiedpushserver/util.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+
+	aerogearv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/aerogear/v1alpha1"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type image struct {
+	defaultImage string
```
camilamacedo86, line 15: IMHO it could be a spec of the CR in order to allow we change it without the need to do a new build. E,g https://github.com/aerogear/mobile-security-service-operator/blob/master/pkg/controller/mobilesecurityservice/deployments.go#L79

pkg/controller/unifiedpushserver/util.go
```diff
@@ -0,0 +1,55 @@
...
+func (i image) image() string {
+	customImage, found := os.LookupEnv(i.envVarName)
+	if found {
+		return customImage
+	}
+	return i.defaultImage
+}
+
+var (
+	proxy       = image{"docker.io/openshift/oauth-proxy:v1.1.0", "OAUTH_PROXY_IMAGE"}
```
camilamacedo86, line 28: IMHO It all should be a spec of the CR since are values that could be need to change and indeed we could change the config in the CR without the need of a new build or code/changes.

camilamacedo86: NOTE: If these ENV VARs are used for all Kind/CRDs, implemented and managed for the operator then could be an ENV VAR of the operator defined in the operator.yaml

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -0,0 +1,211 @@
...
+						IntVal: 4180,
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+func newUnifiedPushServerDeployment(cr *aerogearv1alpha1.UnifiedPushServer) *appsv1.Deployment {
+	labels := map[string]string{
```
camilamacedo86, line 68: It could be in a helper for the CR since will be used for all impl which will create/manage the resources. E.g https://github.com/aerogear/mobile-security-service-operator/blob/master/pkg/controller/mobilesecurityservice/helpers.go#L13

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -100,53 +138,185 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		}
 	} else if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// Pod already exists - don't requeue
-	reqLogger.Info("Skip reconcile: Pod already exists", "Pod.Namespace", found.Namespace, "Pod.Name", found.Name)
-	return reconcile.Result{}, nil
-}
+	postgresqlService, err := newPostgresqlService(instance)
```
camilamacedo86, line 187: IHMO the DB should not be called/used in this way and I'd like to recommend do a CR for the database which should have its own controller as was done [here](https://github.com/aerogear/mobile-security-service-operator/tree/master/pkg/controller). 

## Why? 

Shows that it goes against the design proposed for operator-framework. In my point of view UPS and Database cannot be the same type of object/resource since are not the same project.
 
The App/Service is one CR as the Database is another. Note that each CR means one type of resource. If you follow the internal course materials and/or its docs as the examples found in the operatorHub we will see this design properly in action. 

I'd like to highlight that this approach in my point of view goes against similar ideas and concept as do a design driven by the domain as well and that how much more atomic are the objects/types and implementations better are to keep them maintained.

As you can check [here](https://github.com/operator-framework/operator-sdk/issues/1383) do many CRDs and use one controller according to the teammates of the framework also goes against to design goals for the controller-runtime.

## Examples

**Prometheus**
* https://github.com/coreos/prometheus-operator

**This one is a service and database** ( AquaDatabase CR, AquaServer CR, AquaGateway CR and etc )
* https://operatorhub.io/operator/alpha/aqua-operator.v0.0.1
* https://github.com/aquasecurity/aqua-operator/tree/master/deploy/crds
 
**AWS in HELM**( DynamoDB, ECRRepository and etc .. )
* https://operatorhub.io/operator/alpha/aws-service-operator.v0.0.1
* https://github.com/awslabs/aws-service-operator/tree/master/examples

**Etcd** ( etcd Cluster, etcd Backup, etcd Restore )
* https://github.com/coreos/etcd-operator/tree/master/example
* https://operatorhub.io/operator/clusterwide-alpha/etcdoperator.v0.9.4-clusterwide

Please, feel free to check the operators available in https://operatorhub.io/ 

## Extra reasons and argumentations

May in the future we would like to change it to use a community and/or a Redhat operator for the database then following some extra argumentations.

As it was implemented/designed we are not allowed to:

- Just not install the type DB and use a community operator if we wish
- Remove the DB from the project easily since is not very clear what should be removed or not because all code is impl together. 

NOTE: By using OLM we could define that the CR UPS has as a dependency the CR DB

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: Thanks for the feedback!

I've already got a function here in the util file: https://github.com/aerogear/unifiedpush-operator/pull/6/files#diff-5e52281b0323c385631f2ec4a8e2be64R33

In this case I wanted a slightly different service value so I just created this local variable instead.

pkg/controller/unifiedpushserver/util.go
grdryn: I previously had this conversation earlier in this PR with @wei-lee, and we decided to go this way with env vars in the operator rather than a field in the CR: https://github.com/aerogear/unifiedpush-operator/pull/6/files/5d31078dfe82c3d95518f0ab301a3821c74e3513#discussion_r278063400

pkg/controller/unifiedpushserver/util.go
grdryn: Going to mark this one as resolved as I think it's really the same case as the other comment above at https://github.com/aerogear/unifiedpush-operator/pull/6/files#r282880591

Thanks!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -52,9 +48,53 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// TODO(user): Modify this to be the types you create that are owned by the primary resource
-	// Watch for changes to secondary resource Pods and requeue the owner UnifiedPushServer
-	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+	// Watch for changes to secondary resource Deployment and requeue the owner UnifiedPushServer
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
```
camilamacedo86: To keep easier understand the code and keep it maintained I'd recommend following the approach adopted [here](https://github.com/aerogear/mobile-security-service-operator/tree/master/pkg/controller/mobilesecurityserviceapp)

pkg/controller/unifiedpushserver/unifiedpushserver.go
camilamacedo86: Just to clarifies, the label is specific by CR and the units.go not then because of this in MSS Oper each CR has your own helper.

pkg/controller/unifiedpushserver/util.go
laurafitzgerald: @grdryn I'm going to plagiarize your plagiarization. 👐

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: You might be right about this, but I'm not convinced (at least not yet).

In my opinion, the PostgreSQL server is just an implementation detail. We don't want people to have to configure a PostgreSQL through a CR so that then they can configure UPS to use that in their UPS CR: all they want is UPS, and we don't want them to touch this PostgreSQL themselves.

Yes, in future it would be great if all of the DB management stuff was handled by a PostgreSQL operator, but I'm not convinced that this means we need to have a separate controller & CRD for that in this operator. At that point, I don't think it will be any more work really to remove it from here as it would to remove a separate controller & CRD (also, not sure what would happen to our custom PostgreSQL CRs that people might have created in their projects before then).

@wei-lee @maleck13 do either of you have opinions or thoughts on what's best here, if any?

camilamacedo86: You definitely make a terrific work here 🥇 @grdryn and it indeed can be in another pr, I am just trying to advocating according to my convictions. :-)

IMHO:

- In order to have a CR with DB, we can just use the tool to create the kind and controller and basically cut and paste the code impl  (**really small effort**). We could indeed just copy the impl from MSS Oper since it actually is the same need.
- Do not put all impl in the same CR/Controller doesn't make the things harder just the opposite. At the same way that follows good practices as cohesion and encapsulation do not increase the effort. 
- In MSS Oper we used `make` commands to improve the usability and when we provide it by OLM and/or RHMI, for example, we can just declare the CRs dependency according to the guidelines of the framework adopted. 
- Not follow the framework guidelines can lead us to start to face unexpected scenarios and start to impl workarounds which will make harder put the things in the trail after.
- Note that many impl will be required in the project for we are able to release the first version and indeed things as backup and restore process shows required. 
- Sorry, but I personally cannot agree with the approach of we starting new projects from scratch doing what was not supposed to be done. According to my experience, which is unliked be changed in the future and is very likely lead us to situations where will be hard to improve and/or keep the project maintained because it grew unduly (**highly effort in the future.** 

It is just my IMHO and suggestions, however, it is up to you folks.

wei-lee: I have the same view: we should allow users to configure certain aspects of the db through the same CR (e.g. size of the volume), but I don't really think we need a separate CR and have a separate controller to manage the database, and it is not really the job of the operator for this service.  If we are going to have another CR just for the db, then why not just use a separate db operator instead?

camilamacedo86: Hi @wei-lee,
 
A CR/CRD is a class object definition, so by adding all in the same CR/CRD means that all are part of the same object which means impl the DB as part of the Service which goes against not just the guidelines and good practices but our definitions as well.

The DB is not part of the Service it is required by the Service.

**NOTE:** _Impl the DB as part of the service and will not allow us to remove/change it with a small impact and/or effort consequently will be very unlikely we are able to rip off the DB in the future. :-(_  

Following some comments inline.

>  we should allow users to configure certain aspects of the db through the same CR (e.g. size of the volume),

All Service config and requirements are attributes of the Service object class indeed its requirements/config to use a DB.  In this way, it is done/attended in the MSS Oper.

> If we are going to have another CR just for the db, then why not just use a separate db operator instead?  

Unfortunately, as we spoke before do not exist a PostgreSQL oper which can be used in our projects. However, is highly like that it can be provided by Red Hat soon since for OCP 3.11 we have the PostgreSQL card/image in the catalogue. ( Provide and be responsible to maintain a DB oper shows not be our goal too. )

**NOTE**: _If we realize that is better we have our DB oper then the work to do that will be really easier and small than put all impl together in the same CR._ 

Really thank you for your attention.  
I hope that the above info can clarify your questions.

maleck13: My 2 cents here. 

- ideally Postgres would be managed by another operator and UPS would depend on that operator and create a CR for the postgres it needed. Any additional info / config would be done via that CR
- pragmatically, this is not the case and wont be for some time. So I would be in favour of not exposing Database info in the CR (it is important to think of these as an API, the more you expose the more you have to support and be back compat with). Having a default setup that suits 95% of usecases is a good place to start and then wait for the additional use cases edge cases to filter back from actual use cases.

grdryn: Thanks for the feedback everyone. I'm going to leave it as-is for the moment and merge this today.

For now until there's a postgresql operator that we can call out to, I don't think it's even a good idea to expose the db volume size through the CR, since that might give the impression that it could be tweaked (e.g. someone tries to _increase_ it from 10Gi to 20Gi after a few months, and expects something to happen...).

laurafitzgerald: Hi all, I believe we expose these in the MSS Operator. Should we also change it there? cc @davidffrench @camilamacedo86

davidffrench: @laurafitzgerald You are correct and we also have an individual CR specifically for the DB. This won't be changed for us.

I agree with Craig, what I would like to see is for these values to be optional in our db CR spec with default values defined in the operator when these are not provided.

PR #7, opened 2019-05-14
Add status phase to UPS CRD

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9013

### Verification steps
1. Run the operator (see README changes here)
2. Watch the `status.phase` field. It should be "Provisioning" at the start, then eventually get to "Complete" (once the UPS pod is Ready):

```
watch oc get ups example-unifiedpushserver -o template={{.status.phase}}
```

### DIFF
--- a/pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
+++ b/pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
@@ -18,9 +18,7 @@ type UnifiedPushServerSpec struct {
 // UnifiedPushServerStatus defines the observed state of UnifiedPushServer
 // +k8s:openapi-gen=true
 type UnifiedPushServerStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	Phase StatusPhase `json:"phase"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -43,6 +41,14 @@ type UnifiedPushServerList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []UnifiedPushServer `json:"items"`
 }
+
+type StatusPhase string
+
+var (
+	PhaseEmpty     StatusPhase = ""
+	PhaseComplete  StatusPhase = "Complete"
+	PhaseProvision StatusPhase = "Provisioning"
+)
 
 func init() {
 	SchemeBuilder.Register(&UnifiedPushServer{}, &UnifiedPushServerList{})
--- a/pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
+++ b/pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
@@ -138,6 +138,11 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
+	if instance.Status.Phase == aerogearv1alpha1.PhaseEmpty {
+		instance.Status.Phase = aerogearv1alpha1.PhaseProvision
+		r.client.Status().Update(context.TODO(), instance)
+	}
+
 	persistentVolumeClaim, err := newPostgresqlPersistentVolumeClaim(instance)
 	if err != nil {
 		return reconcile.Result{}, err
@@ -317,6 +322,11 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, nil
 	} else if err != nil {
 		return reconcile.Result{}, err
+	}
+
+	if foundUnifiedpushDeployment.Status.ReadyReplicas > 0 && instance.Status.Phase != aerogearv1alpha1.PhaseComplete {
+		instance.Status.Phase = aerogearv1alpha1.PhaseComplete
+		r.client.Status().Update(context.TODO(), instance)
 	}
 
 	// Resources already exist - don't requeue
--- a/README.adoc
+++ b/README.adoc
@@ -1,2 +1,42 @@
 = UnifiedPush Operator
 
+== Development
+
+=== Prerequisites
+
+- Access to an OpenShift cluster with admin privileges to be able to create Roles.
+  https://github.com/minishift/minishift[Minishift] is suggested.
+
+- Go, Make, dep, operator-sdk, oc, kubectl (kubectl can just be a symlink to oc)
+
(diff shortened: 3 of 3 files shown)


pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
```diff
@@ -21,6 +21,9 @@ type UnifiedPushServerStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	PostgresqlReady  bool `json:"postgresqlReady"`
+	UnifiedpushReady bool `json:"unifiedpushReady"`
```
grdryn: If these are actual status fields that we want, how are the names/structure?

Possible alternative:

```
ready:
  postgresql: true
  unifiedpush: true
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -319,6 +324,11 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
+	if foundUnifiedpushDeployment.Status.ReadyReplicas > 0 && instance.Status.UnifiedpushReady != true {
+		instance.Status.UnifiedpushReady = true
+		r.client.Status().Update(context.TODO(), instance)
+	}
```
grdryn, line 330: As you can see, this block is pretty much the same as the other one on L187 above. I'm not too bothered about the code duplication, but there are a couple of things I'm not sure about:

- Should I re-fetch the contents of the UPS resource (`instance` here) every time before checking (e.g. `instance.Status.UnifiedpushReady != true`)? It's already fetched at the beginning of this function, but maybe should be fetched again?
- Should I leave all of this until the end of the function, so that it's only calling the `r.client.Status().Update(context.TODO(), instance)` once? I guess that's maybe not the best idea, since there are many conditional returns before the end, which might mean that these don't get set promptly (I guess they'd eventually get set once everything else is okay).

deploy/crds/aerogear_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -33,6 +33,15 @@ spec:
...
           type: object
+          properties:
+            postgresqlReady:
+              type: boolean
+              description: 'postgresqlReady is set to boolean true if the PostgreSQL
+                deployment has at least one ready Pod'
+            unifiedpushReady:
+              type: boolean
+              description: 'unifiedpushReady is set to boolean true if the UnifiedPush
+                deployment has at least one ready Pod'
```
grdryn: Not sure how important it is to add this stuff for `status` fields. I see it is usually done for `spec` fields, but I guess it does no harm, and is maybe additional docs.

wei-lee: normally speaking the validation is not required on the `status` field, mainly because it is not set by users. It is more important to provide some docs to explain what values mean in the status field.

pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
wei-lee: normally speaking, this is up to the maintainer to decide what data should be put in the `status` field. But I think there should be at least a single field to indicate if the service/application is ready or not. So I would prefer:

```
status: prepare|provisioning|running|error
```
so that if I need to check if the service is ready to use, I can just check this field. With the current implementation, I have to check both fields but that is very specific to the application itself and I don't really care which component is ready or not.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
wei-lee: > Should I re-fetch the contents of the UPS resource (instance here) every time before checking (e.g. instance.Status.UnifiedpushReady != true)? It's already fetched at the beginning of this function, but maybe should be fetched again?

No, I don't think so. 

> Should I leave all of this until the end of the function, so that it's only calling the r.client.Status().Update(context.TODO(), instance) once? I guess that's maybe not the best idea, since there are many conditional returns before the end, which might mean that these don't get set promptly (I guess they'd eventually get set once everything else is okay).

I don't think it is a hard requirement, especially if the status field is important for the reconcile function itself. If it's only done once at the end, then there is a chance it will never be executed if there is an unexpected error occurred somewhere. I think it is ok to update the instance once there are actually changes made to it, but not to call update if there is no changes made.

pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
grdryn: Cool, thanks! That's what I was mostly unclear about! :)

pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go
```diff
@@ -44,6 +42,14 @@ type UnifiedPushServerList struct {
 	Items           []UnifiedPushServer `json:"items"`
 }
 
+type StatusPhase string
+
+var (
```
grdryn, line 47: I mostly copied these from the keycloak operator. It had a lot more, that don't apply here (at least not yet). I'm happy to change any of them or add/remove as desired.

aliok:
@grdryn I am receiving this compile error when I do `make code/run`:
```
gofmt -w `find . -type f -name '*.go' -not -path "./vendor/*"`
operator-sdk generate k8s
INFO[0000] Running code-generation for Custom Resource group versions: [aerogear:v1alpha1, ] 
Generating deepcopy funcs
INFO[0000] Code-generation complete.                    
go generate ./...
operator-sdk up local --namespace=unifiedpush
INFO[0000] Running the operator locally.                
# github.com/aerogear/unifiedpush-operator/vendor/sigs.k8s.io/yaml
vendor/sigs.k8s.io/yaml/yaml.go:42:48: undefined: DisallowUnknownFields
# github.com/aerogear/unifiedpush-operator/vendor/k8s.io/client-go/transport
vendor/k8s.io/client-go/transport/round_trippers.go:470:9: undefined: strings.Builder
FATA[0009] failed to run operator locally: (exit status 2) 
make: *** [Makefile:14: code/run] Error 1
```
I also did `dep ensure` and tried again, same error.

UPDATE: I think this is related to my local Go version. But don't understand how it worked in previous PR :confused: 
UPDATE #2: Yes, it was the Go version. Operator Framework folks like to live on the bleeding edge :)

grdryn:
@aliok that's interesting...I thought the promise of Go was that nothing new was ever going to be introduced or removed :thinking:

aliok:
Verified now, works just fine. Feel free to merge.

grdryn:
Thanks a lot @aliok!

PR #8, opened 2019-05-15
Handle UPS version updates

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9012

Code is checking the image of container specs (**specs**! not actually checking the existing pods) in the deployment spec and if they are different than the desired image, it updates the deployment to use those images.
This kind of reconciliation code is borrowed from https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L128

Verification:
* Prepare cluster: `make cluster/clean && make cluster/prepare`
* Run the operator: `make code/run`
* Deploy the sample unified push: `oc apply -f deploy/crds/aerogear_v1alpha1_unifiedpushserver_cr.yaml`
* Wait until everything is provisioned
* Edit 1 or more image in deployments (unifiedpush and its postgres)
* You will see that the operator will update the images back

### DIFF
--- a/pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
+++ b/pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
@@ -187,6 +187,26 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		}
 	} else if err != nil {
 		return reconcile.Result{}, err
+	} else {
+		desiredImage := postgresql.image()
+
+		containerSpec := findContainerSpec(foundPostgresqlDeployment, POSTGRES_CONTAINER_NAME)
+		if containerSpec == nil {
+			reqLogger.Info("Skipping image reconcile: Unable to find container spec in deployment", "Deployment.Namespace", foundPostgresqlDeployment.Namespace, "Deployment.Name", foundPostgresqlDeployment.Name, "ContainerSpec", POSTGRES_CONTAINER_NAME)
+		} else if containerSpec.Image != desiredImage {
+			reqLogger.Info("Container spec in deployment is using a different image. Going to update it now.", "Deployment.Namespace", foundPostgresqlDeployment.Namespace, "Deployment.Name", foundPostgresqlDeployment.Name, "ContainerSpec", POSTGRES_CONTAINER_NAME, "ExistingImage", containerSpec.Image, "DesiredImage", desiredImage)
+
+			// update
+			updateContainerSpecImage(foundPostgresqlDeployment, POSTGRES_CONTAINER_NAME, desiredImage)
+
+			// enqueue
+			err = r.client.Update(context.TODO(), foundPostgresqlDeployment)
+			if err != nil {
+				reqLogger.Error(err, "Failed to update Deployment", "Deployment.Namespace", foundPostgresqlDeployment.Namespace, "Deployment.Name", foundPostgresqlDeployment.Name)
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
+		}
 	}
 
 	postgresqlService, err := newPostgresqlService(instance)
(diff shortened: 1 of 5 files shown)


aliok:
@wei-lee  @pb82 any idea around the problem above?

grdryn:
@aliok I wonder if this is because `foundUnifiedpushDeployment` that you're modifying and updating also contains the `status`, and that could have changed since you read it? :thinking:

*Edit:* You could maybe try doing the update with `unifiedpushDeployment` rather than `foundUnifiedpushDeployment`, since that won't have the `status`, and doing an 'update' with that would hopefully be like an 'apply' (if it would work, then it cleans up the logic also, since you don't need to modify the existing `spec.Image` variable in the code I think)

aliok:
@grdryn @wei-lee 
This is ready for review

aliok:
Rebased now

pkg/apis/aerogear/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,5 +1,21 @@
 // +build !ignore_autogenerated
 
+/*
```
grdryn: /me wonders why this never updated for me. What version of the operator-sdk cli do you have?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -187,6 +187,26 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+			// update
+			updateContainerSpecImage(foundPostgresqlDeployment, POSTGRES_CONTAINER_NAME, desiredImage)
+
+			// enqueue
+			err = r.client.Update(context.TODO(), foundPostgresqlDeployment)
+			if err != nil {
+				reqLogger.Error(err, "Failed to update Deployment", "Deployment.Namespace", foundPostgresqlDeployment.Namespace, "Deployment.Name", foundPostgresqlDeployment.Name)
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
```
grdryn, line 208: I guess requeuing here is probably optional, right? What are the advantages of doing it, rather than just continuing?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -187,6 +187,26 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		}
 	} else if err != nil {
 		return reconcile.Result{}, err
+	} else {
```
grdryn, line 190: I think this `else` could be removed (un-nest the contents), what do you think?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -187,6 +187,26 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		desiredImage := postgresql.image()
+
+		containerSpec := findContainerSpec(foundPostgresqlDeployment, POSTGRES_CONTAINER_NAME)
+		if containerSpec == nil {
```
grdryn, line 194: Are there situations where this could happen? :thinking: If so, should we stop progress and requeue or something, rather than just writing an info-level log message?

grdryn (approved):
I've tried this out locally and it worked as expected. I modified the oauth-proxy image for the deployment, and it got set back and this message showed up in the operator logs:

```
{"level":"info","ts":1558007494.8795693,"logger":"controller_unifiedpushserver","msg":"Container spec in deployment is using a different image. Going to update it now.","Request.Namespace":"unifiedpush","Request.Name":"example-unifiedpushserver","Deployment.Namespace":"unifiedpush","Deployment.Name":"example-unifiedpushserver","ContainerSpec":"ups-oauth-proxy","ExistingImage":"quay.io/openshift/origin-oauth-proxy:latest","DesiredImage":"docker.io/openshift/oauth-proxy:v1.1.0"}
```

pkg/apis/aerogear/v1alpha1/zz_generated.deepcopy.go
grdryn: I've got `operator-sdk version: v0.7.0+git`, and when I run `make code/run` (which runs the generation stuff), this get removes. lol

aliok:
```
[aliok@localhost ~]$ operator-sdk --version
operator-sdk version v0.2.1
```

Let me update mine :)

Operator SDK guys are releasing like a boss. New minor release every month. Mine was only 5 months old.

aliok: Ok, now I have the same behavior

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: Frankly, I don't know...
I simply copy pasted that part from https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L138

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: I think no.

In that case, image check would be executed even when there was no deployment and just created in the first `if` block here https://github.com/aerogear/unifiedpush-operator/pull/8/files#diff-4deb70b03698dcadb6ba778ed52ce06bR183

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: I actually didn't know what to do and simply logged a message :grin: 

> Are there situations where this could happen?

Yeah, when somebody changes the container name manually in the deployment. Or, if there's a mismatch in those among different operator versions.

> If so, should we stop progress and requeue or something

What do you mean by this?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: In that case it would be fine either with or without actually, right?

Thinking about when the deployment is created in the `if` block above: do you think maybe we should do a requeue after the create, like you saw in the memcached operator for other spec updates? :thinking: 

**Edit:** Yep, I see that memcached-operator example also does it after creation of objects which we currently don't: https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L123

Maybe we should?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Interesting! I guess it's "well now things are different than when we got triggered, so we should start again". I like that, and there are maybe other places we should do that here too.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Why I mean is that if we enter this block, something has gone terribly wrong that we should deal with, but I'm not 100% sure what to do.

Since the operator is the owner of the deployment, I think it should reconcile to its own state of the world -- if someone else changed something manually, it should change it back. If an earlier version of the operator did something differently, this version should update to its own way (similar to how this PR is making sure it gets updated to the expected images, I think eventually it should update to the expected _everything_).

What I think I'm saying is that I think we should be doing something equivalent to `kubectl apply`, rather than checking for specific fields, and allowing for variation in everything else. Does that make sense? I'm not sure if it's possible, or not. Maybe doing update with the `postgresqlDeployment` object rather than hoping that the `foundPostgresqlDeployment` object has the right structure for us to be able to update the image fields _might_ be possible (I haven't tried).

Again though, this PR works, so is good progress and could be merged without trying to go down that rabbit hole! :smile: 

**Edit:** To give a more concrete example of this: imagine a case where we add a new container to the Pod owned by the Deployment (or rather, owned by the ReplicaSet owned by the Deployment). In the pre-operator world, we could just `kubectl apply -f deployment.yaml` and we'd know that this would bring it up-to-date (both this new container would be added, and any other changes would be applied too). Ideally once this operator is stable, we'd be able to make such changes by just updating with the objects that we instantiate for initial creation, rather than doing that for initial creation and then separately having to add a bunch more logic to check every little thing for this update case.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: cc @pb82 any idea?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: Then instead of checking specific keys in the deployment, we should compare everything.

@grdryn I think what you say makes sense. However, I don't see anything like that in other operators in RHMI. So, not a best practice perhaps. But I cannot tell why.

ping @wei-lee @maleck13
(summary: this PR ensures the image names and versions in the deployment. my code only checks the image field but @grdryn asks why we should not simply compare everything.)

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: @pb82 this could maybe have been an alternative solution to https://github.com/integr8ly/application-monitoring-operator/pull/38 also

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
wei-lee: I think for now just checking the image field is enough, that is the main concern for us for the moment. We can improve this on later on depending on what we see how people using this operator.

aliok: @grdryn gonna rebase and merge this one then

grdryn: Sounds good! :+1:

PR #9, opened 2019-05-15
feat: init e2e test

psturc:
### Verification steps
1. Target some cluster with oc (with cluster-admin privileges)
2. Run
   ```
   export IMAGE="quay.io/aerogear/unifiedpush-operator:AEROGEAR-9147"
   operator-sdk test local ./test/e2e --image $IMAGE --go-test-flags "-v"
   ```

### Comments
* after discussion with @grdryn, I decided to move `build & push` phase from Travis job to Jenkins. I commented out the steps ini [.travis.yml](https://github.com/aerogear/unifiedpush-operator/blob/AEROGEAR-9147/.travis.yml#L26-L35) just for now.


grdryn (changes requested):
I ran the test and it works (and I also tested it with the latest changes from master and it works there too)!

I don't like that when I ran `dep ensure` that I got a different `Gopkg.lock` and new things in `vendor/` though: those should be committed! :smiley: I'm not sure if `dep` has anything equivalent to `devDependencies` in `npm`...it doesn't appear to at first blush.

This will also need a rebase.

Jenkinsfile
```diff
@@ -0,0 +1,113 @@
+#!groovy
```
grdryn: Completely optional, but Declarative Pipeline is nicer than Scripted Pipeline IMHO (even if some of it ends up being nested-scripted anyway)

Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+ADD deploy/operator.yaml /namespaced.yaml
+ADD ${testFileName} /${testFileName}
+"""
+
+node ('operator-sdk') {
+    stage ('Checkout') {
+        checkout scm
+    }
+    stage('Vendor the dependencies') {
+        sh 'dep ensure'
```
grdryn: As mentioned in the review comment, if there are any changes after this, something hasn't been done properly, so maybe failing here would be good (or marking as unstable if you still want to have the test executed).

Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+
+        stage('New project in OpenShift') {
+            openshift.newProject(openshiftProjectName)
+        }
+
+        openshift.withProject(openshiftProjectName) {
+            stage("Build ${operatorName} & ${operatorName}-test binaries") {
+                sh """
+                export GOOS=linux GOARCH=amd64 CGO_ENABLED=0
+                go build -o ${operatorName} cmd/manager/main.go
```
grdryn: Maybe this should execute `make code/compile`, so that it's the same as what we'd run locally? Or alternatively add a `make test/e2e` for the below line that runs the build task first (may also encourage us to run the e2e tests locally more often :wink: )

test/e2e/unifiedpush_test.go
```diff
@@ -0,0 +1,57 @@
...
+var (
+	retryInterval        = time.Second * 5
+	timeout              = time.Second * 200
+	cleanupRetryInterval = time.Second * 1
+	cleanupTimeout       = time.Second * 5
+)
+
+func TestUnifiedpush(t *testing.T) {
+	unifiedpushList := &aerogearv1alpha1.UnifiedPushServerList{
+		TypeMeta: metav1.TypeMeta{
```
grdryn: I think you can remove this `TypeMeta` section. It should be implied by the fact that this is an `&aerogearv1alpha1.UnifiedPushServerList`, and having it set explicitly here could lead to weirdness when we get to `v1beta1` or other versions: where we might get compiler/tooling help to update the versions of objects, these string properties could be missed.

Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+def openshiftProjectName = "test-${operatorName}-${currentBuild.number}-${currentBuild.startTimeInMillis}"
+def operatorDockerImageName = "docker-registry.default.svc:5000/${openshiftProjectName}/${operatorName}-test:latest"
+def testFileName = 'go-test.sh'
+
+def testFileContent = """#!/bin/sh
+${operatorName}-test -test.parallel=1 -test.failfast -root=/ -kubeconfig=incluster -namespacedMan=namespaced.yaml -test.v
+"""
+
+def dockerfileContent = """
+FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:7.6
```
grdryn: Would it be better or worse to have this as a file in the repo, rather than writing it here for every run? My thinking is that it might make it more obvious that if we make changes to the main Dockerfile, that this should probably change too.

Alternatively, what about building the main image first, then use that as the `FROM` in the test image that just adds the test stuff? That way we could have this Jenkinsfile be the thing that builds the images that we push to quay.io: we'd be publishing images that we've run e2e tests against, and we'd have built them on infrastructure that we trust more than travis!

I'm not saying that any of this is stuff that we need to do here in this PR, just ideas for the future! :)

psturc: This is actually a great idea and I didn't realise that we could [push the image built by OpenShift](https://blog.openshift.com/pushing-application-images-to-an-external-registry/) right to quay.io. I will create a new jira for this.

aliok:
@psturc need any help with this PR?

Jenkinsfile
```diff
@@ -0,0 +1,219 @@
...
+        stage("Build operator-test image") {
+            steps{
+                dir("${env.CLONED_REPOSITORY_PATH}") {
+                    script {
+                        def operatorTestDockerfileContent = """
+                        FROM ${env.OPERATOR_CONTAINER_IMAGE_NAME}
+                        ADD build/_output/bin/unifiedpush-operator-test /usr/local/bin/unifiedpush-operator-test
+                        ADD deploy/operator.yaml /namespaced.yaml
+                        ADD build/test-framework/go-test.sh /go-test.sh
+                        """
```
psturc, line 166: @grdryn I couldn't use a Dockerfile for this case since `FROM ${env.OPERATOR_CONTAINER_IMAGE_NAME}` will be always variable and I couldn't use it as an argument, since it's not supported when running binary build.

`WARNING: Specifying build arguments with binary builds is not supported`

test/e2e/unifiedpush_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+var (
+	retryInterval        = time.Second * 5
+	timeout              = time.Second * 200
+	cleanupRetryInterval = time.Second * 1
+	cleanupTimeout       = time.Second * 5
+)
+
+func TestUnifiedpush(t *testing.T) {
+	unifiedpushList := &aerogearv1alpha1.UnifiedPushServerList{}
```
aliok, line 21: I know there are no real tests in this file, but I have the feeling we should have the meta passed as here: https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md#2-register-types-with-framework-scheme

```
unifiedpushList := &aerogearv1alpha1.UnifiedPushServerList{
		TypeMeta: metav1.TypeMeta{
			Kind:       "UnifiedPushServer",
			APIVersion: "aerogear.org/v1alpha1",
		},
	}
```

aliok:
@psturc the dummy test ran fine locally. Can't test Jenkins pipeline though.

.travis.yml
```diff
@@ -23,13 +23,13 @@ jobs:
...
-    script:
-    - export BRANCH=$(if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then echo ${TRAVIS_BRANCH}; else echo ${TRAVIS_PULL_REQUEST_BRANCH}; fi)
-    - export TAG=$(git rev-parse --short ${TRAVIS_COMMIT})
-    - docker login --password "${QUAY_PASSWORD}" --username "${QUAY_USERNAME}" quay.io
-    - make setup/travis
-    - operator-sdk build quay.io/aerogear/${OPERATOR_NAME}:${TAG}
-    - docker push quay.io/aerogear/${OPERATOR_NAME}:${TAG}
-    - docker tag quay.io/aerogear/${OPERATOR_NAME}:${TAG} quay.io/aerogear/${OPERATOR_NAME}:${BRANCH}
-    - docker push quay.io/aerogear/${OPERATOR_NAME}:${BRANCH}
+  # - stage: push
```
grdryn: You can remove this stuff altogether rather than just commenting it out. If we need it back we can always get it from previous commits :)

Jenkinsfile
grdryn: Hmm, shouldn't that variable already be replaced by the literal string at this point by Jenkins, before the openshift Build gets executed? Let me brush up on how strings work in Jenkins pipeline again...

Jenkinsfile
```diff
@@ -0,0 +1,219 @@
...
+                                    "strategy": {
+                                        "dockerStrategy": {
+                                            "dockerfilePath": "build/Dockerfile"
+                                        },
+                                        "type": "Docker"
+                                        },
+                                    "output": {
+                                        "to": {
+                                            "kind": "DockerImage",
+                                            "name": "${env.OPERATOR_CONTAINER_IMAGE_NAME}"
```
grdryn: I think this is probably fine, but just note that it's going to overwrite any existing image for that tag on quay.io without having tested it (since that's later in the pipeline here).

Jenkinsfile
```diff
@@ -0,0 +1,219 @@
...
+        GOPATH = "${env.WORKSPACE}/"
+        PATH = "${env.PATH}:${env.WORKSPACE}/bin"
+        GOOS = "linux"
+        GOARCH = "amd64"
+        CGO_ENABLED = 0
+        OPERATOR_NAME = "unifiedpush-operator"
+        OPERATOR_TEST_NAME = "${env.OPERATOR_NAME}-test"
+        OPENSHIFT_PROJECT_NAME = "test-${env.OPERATOR_NAME}-${currentBuild.number}-${currentBuild.startTimeInMillis}"
+        CLONED_REPOSITORY_PATH = "src/github.com/aerogear/unifiedpush-operator"
+        OPERATOR_CONTAINER_IMAGE_NAME = "quay.io/aerogear/${env.OPERATOR_NAME}:${env.BRANCH_NAME}"
```
grdryn, line 23: If I create a PR from my fork, will `${env.BRANCH_NAME}` have my name in it? I'm just thinking that if you created a PR with the same branch name, that it could overwrite my image, maybe? Or in the case of PRs, is this variable the PR number? I can't remember...

I hope it's the PR number actually, because otherwise I could just create a PR from master of my fork ;D

grdryn (approved):
I've left some minor comments there, but generally it lgtm!

grdryn:
Ooh, one more thing: maybe something could be added to the README.adoc about running this? :smiley:

test/e2e/unifiedpush_test.go
psturc: I removed the `TypeMeta` section from original commit as @grdryn mentioned it could make a confusion when the versions of apiversion will change. See [this comment](https://github.com/aerogear/unifiedpush-operator/pull/9#discussion_r284798859).

Looking now at the other test implementations ([keycloak](https://github.com/integr8ly/keycloak-operator/blob/master/test/e2e/deployment_test.go#L29), [gitea](https://github.com/integr8ly/gitea-operator/blob/master/test/e2e/gitea_test.go#L27), [memcached](https://github.com/operator-framework/operator-sdk-samples/blob/master/memcached-operator/test/e2e/memcached_test.go#L41)), and seeing that `TypeMeta` is always used. Any thoughts on final resolution of this? :)

Jenkinsfile
psturc: Yes, fortunately it's [`PR-$PR_NUMBER`](https://quay.io/repository/aerogear/unifiedpush-operator?tab=tags) :)

test/e2e/unifiedpush_test.go
aliok: leaving it to @grdryn then :)

but recommendation would be doing what memcached operator does

Jenkinsfile
psturc: resolved in [this commit](https://github.com/aerogear/unifiedpush-operator/pull/9/commits/73b7c079ff50c29dbec8533a16297962b4f3622d)

grdryn: LGTM! :+1:

grdryn (approved):

PR #10, opened 2019-05-16
Add ServiceMonitor definition for Prometheus

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9011

Running `make cluster/prepare` will now label the namespace with
`monitoring_key=middleware`, which the prometheus-operator provisioned
by the application-monitoring-operator[1] will look out for.

The prometheus-operator can read the ServiceMonitor CR, and instruct
the Prometheus instance to start scraping metrics as appropriate.

This change doesn't create any GrafanaDashboard or PrometheusRule
CRs. Since we're just exposing the default metrics from an operator,
it should be up to the SRE team to decide what they want to be able to
see and be alerted by, and they would likely want this to be the same
for all operators, rather than what _we_ decide.

[1] https://github.com/integr8ly/application-monitoring-operator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,12 @@ test/unit:
 
 .PHONY: cluster/prepare
 cluster/prepare:
-	-oc new-project $(NAMESPACE)
-	-kubectl create -f deploy/service_account.yaml
-	-kubectl create -f deploy/role.yaml
-	-kubectl create -f deploy/role_binding.yaml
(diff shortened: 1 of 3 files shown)


wei-lee (commented):
the changes looks good. We might want to talk to the SRE team to decide if we need to add dashboards, or if the prefer to do it. But we can add them later.

wei-lee (approved):

PR #13, opened 2019-05-22
Gitignore JetBrains files

aliok:


grdryn (approved):

grdryn:
@aliok note that if you don't want to worry about adding this to every project that you contribute to, you can add a `core.excludesfile` to your `~/.gitconfig` which is like a "user gitignore" rather than "project gitignore". E.g. I've got:

```ini
[core]
        excludesfile = ~/.gitignore
```

aliok:
Thanks for the tip @grdryn

PR #14, opened 2019-05-23
Android variant CRD

aliok:
https://issues.jboss.org/browse/AEROGEAR-9016

See individual commits for explanation (they're gonna be squashed later)

Verification:
* `kubectl apply -f deploy/crds/push_v1alpha1_androidvariant_crd.yaml`
* `kubectl apply -f deploy/crds/push_v1alpha1_androidvariant_cr.yaml`
* `kubectl get androidvariants`


deploy/crds/push_v1alpha1_androidvariant_cr.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: push.aerogear.org/v1alpha1
+kind: AndroidVariant
+metadata:
+  name: example-androidvariant
+spec:
+  # Add fields here
+  variantId: 'deadbeef-badf00d'
```
wei-lee: is this field generated by UPS, or do we expect user to specify the id? If it is generated, then it shouldn't be in the `spec` field, but in `status` field instead as data in `spec` fields are normally entered by users, not generated.

aliok: yeah, see the TODO in the PR description. Doing that change currently.

deploy/crds/push_v1alpha1_androidvariant_cr.yaml
```diff
@@ -0,0 +1,11 @@
...
+  name: 'example-android'
+  description: 'My super Android variant'
+  projectNumber: '123456'
+  googleKey: 'somekeyinbase64=='
```
wei-lee: are `projectNumber` and `googleKey` fields actually match UPS? From UPS the UI asks for `Server Key` and `Sender Id`? Should we keep the same?

aliok: they match the underlying data model name: https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/AndroidVariant.java#L30

don't know the UI. we can make a preference. no strong opinion here.

aliok: renamed fields according to UI @wei-lee

deploy/crds/aerogear_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -9,8 +9,6 @@ spec:
     listKind: UnifiedPushServerList
     plural: unifiedpushservers
     singular: unifiedpushserver
-    shortNames:
-      - ups
```
grdryn: Are short names not good? I prefer `oc get ups` rather than `oc get unifiedpushserver`

deploy/crds/push_v1alpha1_androidvariant_crd.yaml
```diff
@@ -0,0 +1,70 @@
...
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            description:
+              description: Some description of the app.
+              type: string
+            senderId:
+              description: The "Google Project Number" (or Sender ID) from the API
```
grdryn: I notice that the auto-generated fields above follow the go doc standard for their description fields. E.g. "`thing` \<verb> [...]". Should these follow the same pattern?

pkg/apis/push/v1alpha1/androidvariant_types.go
```diff
@@ -0,0 +1,52 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// AndroidVariantSpec defines the desired state of AndroidVariant
+// +k8s:openapi-gen=true
+type AndroidVariantSpec struct {
+	// Some description of the app.
```
grdryn: Same comment as the one about descriptions in the crd file above

grdryn:
@psturc should the Jenkins job be `Required`? IMHO that's more critical than the travis one now.

image

Also, if this was made from Ali's fork rather than this repo, I guess we'd see one travis and one jenkins check, rather than two of each. Does that sound right?

deploy/crds/aerogear_v1alpha1_unifiedpushserver_crd.yaml
aliok: wtf? I didn't delete that one :S

`operator-sdk generate openapi` deleted it probably. gonna have a look.

deploy/crds/push_v1alpha1_androidvariant_crd.yaml
aliok: I took those descriptions from UPS code: https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/AndroidVariant.java#L42

Can be improved of course.

grdryn: I'd suggest maybe something like:

> senderId is the "Google Project Number" [...]"

wdyt?

psturc:
@grdryn 

> should the Jenkins job be Required? IMHO that's more critical than the travis one now.

definitely, I just set it up

> Also, if this was made from Ali's fork rather than this repo, I guess we'd see one travis and one jenkins check, rather than two of each. Does that sound right?

yes. see [this PR](https://github.com/aerogear/unifiedpush-operator/pull/15)

aliok:
@grdryn pushed a new commit

grdryn (approved):
LGTM!

Did you find out what caused the removal of the shortname from the other crd on your end?

deploy/crds/push_v1alpha1_androidvariant_crd.yaml
```diff
@@ -0,0 +1,70 @@
...
+            ready:
+              type: boolean
+            variantId:
+              type: string
+          required:
+          - ready
+          type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
```
grdryn, line 68: I'm not sure if this question is valid, asking anyway! :) Did the older ups-config-operator have CRDs, and if so was there one called AndroidVariant?

aliok: it was using the binding mechanism with secrets. No CRDs!

aliok:
@grdryn 

> Did you find out what caused the removal of the shortname from the other crd on your end?
Yeah try yourself: `operator-sdk generate openapi`.

I think the shortname should be somehow defined somewhere in `pkg/apis/aerogear/v1alpha1/unifiedpushserver_types.go`

PR #16, opened 2019-05-24
IOS variant CRD

aliok:
https://issues.jboss.org/browse/AEROGEAR-9017

Verification:

```
k apply -f deploy/crds/push_v1alpha1_iosvariant_crd.yaml
k apply -f deploy/crds/push_v1alpha1_iosvariant_cr.yaml
k get iosvariants
```

Expectedoutput:
```
NAME                 PRODUCTION   VARIANTID   READY
example-iosvariant   false   
```


deploy/crds/push_v1alpha1_iosvariant_cr.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: push.aerogear.org/v1alpha1
+kind: IOSVariant
+metadata:
+  name: example-iosvariant
+spec:
+  description: 'My super iOS variant'
+  certificate: 'somekeyinbase64=='
+  passphrase: '123456'
```
grdryn, line 8: Should this maybe be base64 encoded also? I'm not sure what characters are allowed to be in the passphrase, and/or if any of those allowed characters could cause trouble here.

grdryn (approved):

grdryn:
I think maybe we might need to eventually have a field for the name of the UnifiedPushServer CR that corresponds to the server where the variant should be created. I don't think there's anything preventing a user from creating 2 or more UnifiedPushServer CRs resulting in 2 or more servers being managed by the operator. In that case, we should be specifying which one a given variant CR should be tied to, right?

We can add that if necessary once we get to the implementation of the variant controller.

deploy/crds/push_v1alpha1_iosvariant_cr.yaml
aliok: Yeah, that's right. I just typed something there as the operator is not reading the CR at the moment. I suggest we update it later.

aliok:
@grdryn 

> I think maybe we might need to eventually have a field for the name of the UnifiedPushServer CR that corresponds to the server where the variant should be created. I don't think there's anything preventing a user from creating 2 or more UnifiedPushServer CRs resulting in 2 or more servers being managed by the operator. In that case, we should be specifying which one a given variant CR should be tied to, right?

This is a good point. We need to clarify how UPS will be used. Multitenant or per-customer/per-project?

If there will be only 1 UPS instance perr RHMI installation, then the CRs would be created per user namespace. In that case we won't worry about that problem.

If users will have the ability to provision their own UPS instance in their namespace, we don't also care about that problem because it is their responsibility.

Perhaps I am misreading the situation or underestimating something.
Overall this is a confusing subject and I remember some things were decided to be used multitenant during the workshop in April(or March) and some were not. 

cc @wei-lee

aliok:
@wei-lee merging this PR but the question above is still valid.

wei-lee:
@aliok @grdryn In RHMI it is single-tenanted, so there will be only one managed UPS instance. This managed UPS instance will only watch MDC namespace for CRs. Developers maybe able to provision their own UPS instances into their own namespace, however, it's unlikely the UPS server they provisioned will be able to read CRs outside their own namespaces, simply because they will not have the permission to grant their own UPS the permission to read resources from other namespaces.

I don't think it is necessary to add the ups server name to the variant CR. Instead, it might be ok to add the url of the UPS console to the CR, or as a configmap. This url is likely to be used by user to access the console, rather than the server name.

Either way, it is ok to record some information about the server in the CR if we want to, but that should really for user information only. To prevent a CR from being consumed by more than 1 operators at the same time is not a problem for the operator to solve, instead, it should use k8s permission system.

grdryn:
@wei-lee @aliok yeah, my thinking is more around the abilities of operators more generally (i.e. when we eventually get it to the operator hub), rather than the integreatly use case. You're right that we should be getting our immediate use case working well first :+1: 

To think about it in more simple terms than several UPS instances in different namespaces, or several operator instances each managing one UPS -- consider the case of a namespaced UPS operator that just watches for UnifiedPushServer and IOSVariant/AndroidVariant CRs in its own namespace: someone could create 2 UnifiedPushServer CRs, just like they can create many IOSVariant ones. In this case, the operator will legitimately provision 2 distinct UPS instances in the same namespace, and now which one should each IOSVariant be associated with?

Yeah, it's kind of a _"why would you want to do that?"_ kind of case, but I think it's an appropriate way to think about how operators work (rather than how UPS works).

wei-lee:
@grdryn > someone could create 2 UnifiedPushServer CRs, just like they can create many IOSVariant ones. In this case, the operator will legitimately provision 2 distinct UPS instances in the same namespace, and now which one should each IOSVariant be associated with?

I don't think we should support this. The UPS operator itself should know if there is an instance of UPS already provisioned and if so it should not allow provisioning another one.

grdryn:
@wei-lee I like the idea of not "supporting" that use case, at least in the beginning.

I'm conflicted around the idea of the operator refusing to provision a second instance though. The second instance may specify a completely different configuration or something.

If we really wanted to limit to one instance of UPS per operator, I think we shouldn't expose creation of UPS instances with CRs, and instead just immediately provision when the operator starts. That would probably come with its own set of complications though.

wei-lee:
@grdryn we can support the following usecases:

1. The operator will only provision a UPS instance in a namespace if it's not there.
2. If the UPS is already there, and the configurations do not match the CR, then the operator can update the UPS instance to match the CR.

I think this is how most of other operators work. I think at least at the beginning, let's not worry about provisioning multiple instances in the same namespace. If we actually see there are this kind of use cases, then we can discuss if we need to support it.

grdryn:
@wei-lee yeah, I agree, let's not worry about it for the moment. I won't do anything about any of it. The conversation is good though, to make sure we're all thinking in the same direction! :smiley: 

> If the UPS is already there, and the configurations do not match the CR, then the operator can update the UPS instance to match the CR.

I think this would go against people's intuitions about resources: if they want to update the pre-existing UPS configuration, then they should be editing the CR for that instance (rather than a new CR with a different name causing a different one to be modified). I think it would be better to mark the new one as failed without provisioning anything from it.

wei-lee:
@grdryn That is what I meant: modify the instance to match the CR. For new CRs, it should just fail.

PR #18, opened 2019-05-28
Add PushApplication CRD

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9018

Based on the API for PushApplications in UPS, at e.g.
https://github.com/aerogear/aerogear-unifiedpush-server/blob/ff5ba2cb8446d929e5fd426bcb645299b4559914/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/PushApplication.java


pkg/apis/push/v1alpha1/pushapplication_types.go
```diff
@@ -0,0 +1,63 @@
...
+type PushApplicationSpec struct {
+
+	// Description is a description of the app to be displayed in
+	// the UnifiedPush Server admin UI
+	Description string `json:description`
+
+	// Developer is a string containing the name of the developer
+	// that created the app. It is displayed in the UnifiedPush
+	// Server admin UI
+	Developer string `json:developer`
```
aliok: @secondsun is this field optional --> https://github.com/aerogear/aerogear-unifiedpush-server/blob/ff5ba2cb8446d929e5fd426bcb645299b4559914/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/PushApplication.java#L48 ?

aliok (approved):

pkg/apis/push/v1alpha1/pushapplication_types.go
grdryn: It looks like when creating an application from the UI currently, that it's not exposed and a default of `developer` is set:

image

I haven't looked into whether or not that's the UI setting that or if it's a default in the API. I guess the UI that we present in OpenShift might just set a default since it might be difficult to get the openshift user name?

I'm happy to remove it from here if you think that's best, or make it optional here?

aliok: Let's remove it! It would mean less confusion if we don't have these kind of rarely used stuff.

grdryn: Ok cool, I'll remove it now so. We can add it as optional in future if we want to :+1:

How do the other fields look to you?

grdryn:
@aliok another thing: we've now got two api groups in this operator: `aerogear.org` for the UPS CRD, and `push.aerogear.org` for the variants and this pushapplication CRD.

Is that okay, or should I move the UPS CRD into `push.aerogear.org`?

pkg/apis/push/v1alpha1/pushapplication_types.go
aliok: Didn't see anything strange :)

aliok:
@grdryn 

> Is that okay, or should I move the UPS CRD into push.aerogear.org?

Yeah, good point. The reason I introduced `push.a.o` without noticing is basically in JIRA it was suggested to use that. But, I think it it might be a better idea to use a single api group within this operator's resources.

grdryn:
@aliok ok cool. I'll merge this PR as-is for now, because I want to submit another with the start of the client stuff, to get your feedback on that. We can merge the two groups separately :+1: 

Thanks for the review! :pray:

PR #19, opened 2019-05-29
Add initial UPS client package

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9018

NOTE: This package is not currently intended for use outside this
operator project, and is subject to change without notice!


pkg/unifiedpush/client.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// UnifiedpushClient is a client to enable easy interaction with a UPS
+// server
+type UnifiedpushClient struct {
+	url string
+}
+
+// CreateApplication creates an application in UPS
+func (c UnifiedpushClient) CreateApplication(p *pushv1alpha1.PushApplication) (string, string, error) {
```
grdryn, line 56: You can see from the return value of these exposed functions, they're currently returning strings, and I'm not sure if that's the best idea. I wasn't sure whether this client should be modifying the k8s CR that it receives as a pointer to as a param, or if it should just pass back the information so that the controller can update the object itself. Thoughts?

pkg/unifiedpush/client.go
```diff
@@ -0,0 +1,197 @@
...
+
+	var createdApplication pushApplication
+	b, _ := ioutil.ReadAll(resp.Body)
+	json.Unmarshal(b, &createdApplication)
+
+	return createdApplication.ApplicationId, createdApplication.MasterSecret, nil
+}
+
+// CreateVariant creates a Variant on an Application in UPS
+func (c UnifiedpushClient) CreateVariant(o *runtime.Object) (string, error) {
```
grdryn: I'm not sure if this single `CreateVariant` method is the best idea. Should I remove it and instead just expose the (currently internal `createIosVariant` and `createAndroidVariant` methods? Those have slightly better typing I think (currently this will compile if the caller passes it any old thing that seems to implement `runtime.Object`).

grdryn (commented):
@aliok I haven't actually tried any of this out yet, but I figured by making the PR you'd be able to get an idea of the direction I've been going with it, and be able to provide feedback! :)

I'll look into adding some tests, but also please see the review comments here where I've got more questions.

pkg/unifiedpush/client.go
```diff
@@ -0,0 +1,197 @@
...
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := doUPSRequest(req)
+	if err != nil {
+		return "", "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 201 {
```
grdryn, line 79: I tried to pull as much common stuff as possible out into that `doUPSRequest` function that's called above, but I'm not sure how much makes sense.

We could consider putting this check into it, but I'm not sure if _all_ requests would expect a `201` status code response (the current "create" ones do, but we could add more read/update/delete ones).

I think putting this (or some more general check, such as "make sure the statuscode is less than 400") would allow us to pull more stuff into it, such as the reading (and closing) of the response body, and just return the byte slice from it, that we can just `json.unmarshal` here.

Again, I don't know if that's better or worse...opinions welcome!

pkg/unifiedpush/client.go
aliok: IMO, as a convention, it is better to return structs instead of strings.
Passing pointer to the method and making rest client modifying a CR struct kinda sounds strange.
I am not the Golang authority though :)

aliok: On a second thought, when you want to return structs, you need to create new struct types just for the return value. In that case, probably the passing a pointer to the CR and modifying that is good.

pkg/unifiedpush/client.go
aliok: IMO let's leave it that way for now. We can do that move later.

pkg/unifiedpush/client.go
aliok: IMO, exposing createVariant and createAndroidVariant is better.

The type check in this method should not be the responsibility of this package.

pkg/unifiedpush/client.go
aliok: My final idea: leave it as is! :grin: 

Let's see where it brings us.

grdryn:
@aliok The current state of this is that it works for creating PushApplications and AndroidVariants, but UPS currently responds with 400 when I try to create IOSVariants.

Also, to get it to work, an additional service needs to be added, pointing to the UPS container port in the pod rather than the oauth-proxy port, so that the operator doesn't need to go through the oauth flow

aliok:
@grdryn 
> Also, to get it to work, an additional service needs to be added, pointing to the UPS container port in the pod rather than the oauth-proxy port, so that the operator doesn't need to go through the oauth flow

Can you create a ticket for that please?

grdryn:
@aliok sure, here it is: https://issues.jboss.org/browse/AEROGEAR-9300 (I've added it to this epic and sprint also).

In other news, I've got a feeling that the reason my IOSVariant requests might be rejected is because I'm sending it garbage data :see_no_evil:

aliok:
> In other news, I've got a feeling that the reason my IOSVariant requests might be rejected is because I'm sending it garbage data see_no_evil

Oh, yeah :) It is also the same when we were testing UPS admin UI or the MDC. You need to send proper data. I will send you a key and a password privately.

aliok (approved):

PR #22, opened 2019-06-05
AEROGEAR-9160 - Wait for PostgreSQL to be ready before provisioning UPS

jomrazek:
Ready for testing

### DIFF
--- a/pkg/controller/unifiedpushserver/unifiedpushserver.go
+++ b/pkg/controller/unifiedpushserver/unifiedpushserver.go
@@ -91,6 +91,24 @@ func newUnifiedPushServerDeployment(cr *pushv1alpha1.UnifiedPushServer) (*appsv1
 				},
 				Spec: corev1.PodSpec{
 					ServiceAccountName: cr.Name,
+					InitContainers: []corev1.Container{
+						{
+							Name:            POSTGRES_CONTAINER_NAME,
+							Image:           postgresql.image(),
+							ImagePullPolicy: corev1.PullAlways,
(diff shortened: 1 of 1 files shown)


pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -91,6 +91,18 @@ func newUnifiedPushServerDeployment(cr *pushv1alpha1.UnifiedPushServer) (*appsv1
...
 					ServiceAccountName: cr.Name,
+					InitContainers: []corev1.Container{
+						{
+							Name:            POSTGRES_CONTAINER_NAME,
+							Image:           postgresql.image(),
+							ImagePullPolicy: corev1.PullAlways,
+							Command: []string{
+								"/bin/sh",
+								"-c",
+								"until nslookup example-unifiedpushserver-postgresql; do echo waiting for service; sleep 2; done;",
```
grdryn: Unfortunately `nslookup` won't be enough here, since that should return successfully once the `Service` is created. What we care about here though is that there's anything ready for requests behind that `Service` (i.e. Pods that have been marked as Ready that match the selector in the service).

I've just tested that it's not enough also, by scaling down the Postgresql to 0 pods, and then recreating the UPS pod: it passes the initialization phase straight away, even though what we want is for it to fail.

jomrazek: :+1: thanks for testing

grdryn (approved):
LGTM! :100:

PR #24, opened 2019-06-11
fix: instructions in readme

psturc:
### Motivation
CR is created in a different namespace when "unifiedpush" namespace is not targeted (via `oc` or `kubectl`)

Hey @grdryn I'm not sure if this is the best way how to fix the instructions.
Other way could be to [target the namespace with kubectl command](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#setting-the-namespace-preference) but that way it wouldn't be [compatible with `oc`](https://github.com/aerogear/unifiedpush-operator/compare/master...fix-readme?quick_pull=1#diff-7bd4a925c695c2eb0eced3872b9b965fL126)
wdyt?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -143,13 +143,13 @@ make code/run
 3. Create a UPS instance (in another terminal):
 
 ....
-kubectl apply -f deploy/crds/push_v1alpha1_unifiedpushserver_cr.yaml
+kubectl apply -f deploy/crds/push_v1alpha1_unifiedpushserver_cr.yaml -n unifiedpush
 ....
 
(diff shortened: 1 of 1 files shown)


grdryn (approved):
This way looks good @psturc, it's the same way that I fixed it in the Makefile, but clearly forgot about the readme: https://github.com/aerogear/unifiedpush-operator/blob/master/Makefile#L44..L54

Merge away! :100:

grdryn:
Merging this so that I can rebase on top of it for my PR.

PR #27, opened 2019-06-18
[AEROGEAR-9420] Make operator able to listen for CRs in the MDC namespace

jomrazek:


Makefile
```diff
@@ -17,7 +17,7 @@ code/compile:
 
 .PHONY: code/run
 code/run: code/gen
-	operator-sdk up local --namespace=$(NAMESPACE)
+	operator-sdk up local
```
grdryn, line 20: Is it required to remove the `--namespace` here? Is it better without it?

deploy/operator.yaml
```diff
@@ -25,6 +25,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: APP_NAMESPACES
+                value: "unifiedpush"
```
grdryn: Hmm, this doesn't look like it does what it says in the commit message :wink: If we put the mdc namespace name (I'm not sure what it is) in here by default, would things break if it's not there?

psturc:
```suggestion
              value: "unifiedpush"
```
so it looks like the test is failing because of these 2 extra spaces

grdryn (commented):
I've left a couple of questions inline. Also, would you be willing to rebase your commit on top of the latest master, rather than merging the master branch in? It will make the git log cleaner.

Also, it looks like this breaks the Jenkins build. Maybe @psturc can help point you in the right direction with that.

Makefile
jomrazek: --namespace string        The namespace where the operator watches for changes.
We want to watch multiple namespaces or am I wrong? (mdc, ups)?

grdryn: Ah yeah, sorry, I thought that flag had a different purpose. Sounds good! :+1:

grdryn:
I re-triggered the jenkins job in case it was a flake

psturc:
IMO the issue with test now is that the operator is restricted [only for namespace `unifiedpush`](https://github.com/aerogear/unifiedpush-operator/pull/27/files#diff-5e5f804a8ee2fb8c81711b728eb5a21dR29), but the test is [creating a different project name](https://github.com/aerogear/unifiedpush-operator/blob/master/Jenkinsfile#L22).

Since the tests are running only on temporary OpenShift cluster (`oc cluster up`), I think that we don't need to specify a unique name for the projects. @grdryn wdyt?

After changing the project name in Jenkinsfile the test should pass.

grdryn:
> Since the tests are running only on temporary OpenShift cluster (oc cluster up), I think that we don't need to specify a unique name for the projects. @grdryn wdyt?

Is there ever a chance that more than one job run will try to run on the same temporary openshift, and try to use the same "unifiedpush" namespace?

aliok:
I am happy to give this PR a try on a cluster. Let me know when I can try it.

aliok:
Sorry, closed by mistake

deploy/operator.yaml
jomrazek: It doesn't need to be here at all, you're right, I got confused by some samples.

psturc:
@grdryn If we don't increase the number of executors for the CI node (default is 1), then there is no chance that one node with the same OpenShift will be ever used by more than 1 job at the same time. So I think it's fine.

psturc:
Test passed after I [changed the project name to `unifiedpush`](https://mobile-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/unifiedpush-operator/job/PR-27/7/replay/diff). If @grdryn has no objections, I would suggest to make the change in Jenkinsfile then

aliok:
trying to verify this now

aliok (approved):
I verified the PR locally. Created CRs in random namespaces and the operator was able to pick them up.

aliok:
@grdryn @psturc can we merge this already? it will allow MDC team to work on the integration

psturc:
I'm ok with merging this guy once Jenkinsfile is fixed ([this change is mandatory](https://mobile-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/unifiedpush-operator/job/PR-27/7/replay/diff), otherwise the test will be failing in future PRs)

aliok:
@psturc and I decided to merge this one and create a separate PR to fix the Jenkinsfile problem.

PR #28, opened 2019-06-21
AEROGEAR-9434 use ImageStreams

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9434

Use ImageStreams for UPS and its deps (OauthProxy, Postgres).
Also switched from plain Kube Deployments to DeploymentConfigs.

Verification:
* You can deploy the prebuilt image: aliok/ups-opr-jun25-4
* Deploy the operator
* Create unifiedpush CR
* Check the created DeploymentConfigs, Pods, ImageStreams in OpenShift Console
* Delete the CR
* Check if all the resources above are deleted.

Gonna squash my commits after approval.

### DIFF
--- a/deploy/role_binding.yaml
+++ b/deploy/role_binding.yaml
@@ -10,16 +10,3 @@ roleRef:
   kind: ClusterRole
   name: unifiedpush-operator
   apiGroup: rbac.authorization.k8s.io
----
-kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: unifiedpush-operator-image-import
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: system:openshift:controller:image-import-controller
-subjects:
-  - apiGroup: rbac.authorization.k8s.io
-    kind: Group
-    name: system:masters
--- a/deploy/role_binding.yaml
+++ b/deploy/role_binding.yaml
@@ -10,3 +10,16 @@ roleRef:
   kind: ClusterRole
   name: unifiedpush-operator
   apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: unifiedpush-operator-image-import
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:openshift:controller:image-import-controller
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: system:masters
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -9,10 +9,12 @@ type Config struct {
 
 	UPSImageStreamName        string
 	UPSImageStreamTag         string
-	PostgresImageStreamName   string
-	PostgresImageStreamTag    string
 	OauthProxyImageStreamName string
 	OauthProxyImageStreamTag  string
+
+	PostgresImageStreamNamespace string
+	PostgresImageStreamName      string
+	PostgresImageStreamTag       string
 
 	UPSImageStreamInitialImage        string
 	PostgresImageStreamInitialImage   string
@@ -27,14 +29,15 @@ func New() Config {
 
 		UPSImageStreamName:        getEnv("UPS_IMAGE_STREAM_NAME", "ups-imagestream"),
 		UPSImageStreamTag:         getEnv("UPS_IMAGE_STREAM_TAG", "latest"),
-		PostgresImageStreamName:   getEnv("POSTGRES_IMAGE_STREAM_NAME", "ups-postgres-imagestream"),
-		PostgresImageStreamTag:    getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
 		OauthProxyImageStreamName: getEnv("OAUTH_PROXY_IMAGE_STREAM_NAME", "ups-oauth-proxy-imagestream"),
 		OauthProxyImageStreamTag:  getEnv("OAUTH_PROXY_IMAGE_STREAM_TAG", "latest"),
 
+		PostgresImageStreamNamespace: getEnv("POSTGRES_IMAGE_STREAM_NAMESPACE", "openshift"),
+		PostgresImageStreamName:      getEnv("POSTGRES_IMAGE_STREAM_NAME", "postgresql"),
+		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
+
 		// these are used when the image stream does not exist and created for the first time by the operator
 		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
-		PostgresImageStreamInitialImage:   getEnv("POSTGRES_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/centos/postgresql-96-centos7:9.6"),
 		OauthProxyImageStreamInitialImage: getEnv("OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/openshift/oauth-proxy:v1.1.0"),
 	}
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,48 @@
+package config
+
+import "os"
+
+type Config struct {
+	UPSContainerName        string
+	PostgresContainerName   string
+	OauthProxyContainerName string
+
+	UPSImageStreamName        string
+	UPSImageStreamTag         string
+	PostgresImageStreamName   string
+	PostgresImageStreamTag    string
+	OauthProxyImageStreamName string
+	OauthProxyImageStreamTag  string
+
+	UPSImageStreamInitialImage        string
+	PostgresImageStreamInitialImage   string
+	OauthProxyImageStreamInitialImage string
+}
+
+func New() Config {
+	return Config{
+		UPSContainerName:        getEnv("UPS_CONTAINER_NAME", "ups"),
+		PostgresContainerName:   getEnv("POSTGRES_CONTAINER_NAME", "postgresql"),
+		OauthProxyContainerName: getEnv("OAUTH_PROXY_CONTAINER_NAME", "ups-oauth-proxy"),
+
+		UPSImageStreamName:        getEnv("UPS_IMAGE_STREAM_NAME", "ups-imagestream"),
+		UPSImageStreamTag:         getEnv("UPS_IMAGE_STREAM_TAG", "latest"),
+		PostgresImageStreamName:   getEnv("POSTGRES_IMAGE_STREAM_NAME", "ups-postgres-imagestream"),
+		PostgresImageStreamTag:    getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
+		OauthProxyImageStreamName: getEnv("OAUTH_PROXY_IMAGE_STREAM_NAME", "ups-oauth-proxy-imagestream"),
+		OauthProxyImageStreamTag:  getEnv("OAUTH_PROXY_IMAGE_STREAM_TAG", "latest"),
+
+		// these are used when the image stream does not exist and created for the first time by the operator
+		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
+		PostgresImageStreamInitialImage:   getEnv("POSTGRES_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/centos/postgresql-96-centos7:9.6"),
+		OauthProxyImageStreamInitialImage: getEnv("OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/openshift/oauth-proxy:v1.1.0"),
+	}
+}
+
+func getEnv(key string, defaultVal string) string {
+	if value, exists := os.LookupEnv(key); exists {
+		return value
+	}
+
+	return defaultVal
+}
--- a/pkg/controller/unifiedpushserver/postgresql.go
+++ b/pkg/controller/unifiedpushserver/postgresql.go
@@ -5,8 +5,6 @@ import (
 
 	pushv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
 	openshiftappsv1 "github.com/openshift/api/apps/v1"
-	imagev1 "github.com/openshift/api/image/v1"
-
 	"github.com/pkg/errors"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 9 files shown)


pkg/config/config.go
```diff
@@ -0,0 +1,48 @@
+package config
```
aliok, line 1: created this file and the package to hold the config for the operator.

I wanted operator to receive things like image stream details etc. via env vars. In the end they were too many and decided to create this single file which holds all config.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,50 +212,59 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		err = r.client.Create(context.TODO(), postgresqlDeploymentConfig)
+		if err != nil {
+			return reconcile.Result{}, err
 		}
+	} else if err != nil {
+		return reconcile.Result{}, err
 	}
+	//#endregion
 
+	//#region Postgres Service
```
aliok, line 241: These region blocks look really good on Intellij and VSCode (with extension).
They're also code regular comments specifying the block start and end.

Screenshot from 2019-06-21 15-07-48

aliok: @grdryn don't know about Emacs though!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -375,70 +409,83 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
 		}
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+	//#endregion
 
-		proxyContainerSpec := findContainerSpec(foundUnifiedpushDeployment, OAUTH_PROXY_CONTAINER_NAME)
-		if proxyContainerSpec == nil {
-			reqLogger.Info("Skipping image reconcile: Unable to find container spec in deployment", "Deployment.Namespace", foundUnifiedpushDeployment.Namespace, "Deployment.Name", foundUnifiedpushDeployment.Name, "ContainerSpec", OAUTH_PROXY_CONTAINER_NAME)
-		} else if proxyContainerSpec.Image != desiredProxyImage {
```
aliok, line 425: Image reconciliation done with https://issues.jboss.org/browse/AEROGEAR-9012 doesn't make sense anymore with ImageStreams.
Got rid of this block.

aliok:
something failed here, but I can't login to that Jenkins instance. can you create a user for me @psturc ?

deploy/role_binding.yaml
```diff
@@ -10,3 +10,16 @@ roleRef:
   kind: ClusterRole
   name: unifiedpush-operator
   apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRoleBinding
```
grdryn: Could you explain what's going on with this ClusterRoleBinding, it seems odd that we'd be adding a cluster role to a group that isn't our operator

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -75,26 +100,63 @@ func newUnifiedPushServerDeployment(cr *pushv1alpha1.UnifiedPushServer) (*appsv1
...
+					},
+				},
+				openshiftappsv1.DeploymentTriggerPolicy{
+					Type: openshiftappsv1.DeploymentTriggerOnImageChange,
+					ImageChangeParams: &openshiftappsv1.DeploymentTriggerImageChangeParams{
+						Automatic:      true,
+						ContainerNames: []string{cfg.PostgresContainerName},
+						From: corev1.ObjectReference{
+							Kind: "ImageStreamTag",
+							Name: cfg.PostgresImageStreamName + ":" + cfg.PostgresImageStreamTag,
```
grdryn: Is it correct to trigger this UPS DeploymentConfig on an updated PostgreSQL ImageStreamTag update, or should this one instead be in the PostgreSQL DeploymentConfig (I notice there aren't any triggers in there)?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -24,6 +25,10 @@ import (
 
 var log = logf.Log.WithName("controller_unifiedpushserver")
 
+var (
```
grdryn, line 26: Would it make sense to include the `log` variable on the line above in this `var ()` block?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Cool! :+1:

pkg/controller/unifiedpushserver/postgresql.go
```diff
@@ -181,3 +181,27 @@ func newPostgresqlService(cr *pushv1alpha1.UnifiedPushServer) (*corev1.Service,
 		},
 	}, nil
 }
+
+func newPostgresImageStream(cr *pushv1alpha1.UnifiedPushServer) (*imagev1.ImageStream, error) {
```
grdryn: There should already an ImageStream for PostgreSQL in the `openshift` namespace on OpenShift, so we probably don't need to create our own.

pkg/controller/unifiedpushserver/postgresql.go
```diff
@@ -181,3 +181,27 @@ func newPostgresqlService(cr *pushv1alpha1.UnifiedPushServer) (*corev1.Service,
...
+		Spec: imagev1.ImageStreamSpec{
+			Tags: []imagev1.TagReference{
+				{
+					Name: cfg.PostgresImageStreamTag,
+					From: &corev1.ObjectReference{
+						Kind: "DockerImage",
+						Name: cfg.PostgresImageStreamInitialImage,
+					},
+					ImportPolicy: imagev1.TagImportPolicy{
+						Scheduled: true,
```
grdryn: According to @maleck13, we shouldn't set this scheduled part (for now at least). Image imports should instead be done manually during a scheduled maintenance window.

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -64,7 +65,31 @@ func newOauthProxyRoute(cr *pushv1alpha1.UnifiedPushServer) (*routev1.Route, err
 	}, nil
 }
 
-func newUnifiedPushServerDeployment(cr *pushv1alpha1.UnifiedPushServer) (*appsv1.Deployment, error) {
+func newOauthProxyImageStream(cr *pushv1alpha1.UnifiedPushServer) (*imagev1.ImageStream, error) {
```
grdryn, line 68: Like PostgreSQL, there's also a default ImageStream for the oauth-proxy available through the `openshift` namespace, so maybe we could/should use that.

@maleck13 note that for the default one in this case in the openshift project, the only tag (`v1.1.0`) has `importPolicy.scheduled = true` (postgresql doesn't have this).

pkg/controller/unifiedpushserver/postgresql.go
aliok: I thought about it. 
I guess this depends how the ops manage this situation. Perhaps the Postgres version they want to use for UPS is different than the version they want to use in some other place.
If this is nonsense, I am ok  to drop the ImageStream creation here. Same goes to Oauth Proxy.

pkg/controller/unifiedpushserver/postgresql.go
aliok: Ok, will apply that to all image streams we create here

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: > Is it correct to trigger this UPS DeploymentConfig on an updated PostgreSQL ImageStreamTag update

That's a preference IMO. Happy to delete that trigger for UPS.

> or should this one instead be in the PostgreSQL DeploymentConfig 

I will check it. Perhaps I forgot it.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: +1

deploy/role_binding.yaml
aliok: I copied it from Apicurio operator. 

@odra any idea?
@grdryn sounds right.

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: > Is it correct to trigger this UPS DeploymentConfig on an updated PostgreSQL ImageStreamTag update

Sent an email to AeroGear ML about this: https://groups.google.com/forum/#!topic/aerogear/D7vncataJac

pkg/controller/unifiedpushserver/postgresql.go
aliok: Sent an email to AeroGear ML about this: https://groups.google.com/forum/#!topic/aerogear/D7vncataJac

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: > Is it correct to trigger this UPS DeploymentConfig on an updated PostgreSQL ImageStreamTag update

When I don't have this here, the init container fails to resolve the image from the imageStream. I guess this would be possible but I couldn't figure out how. Let's pair on this

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: @grdryn I don't see that on Minishift.

I don't have a OCP cluster handy but if it is available in OCP and not in OKD by default, perhaps we can check the existence of the image stream first. What do you think?

Screenshot from 2019-06-25 10-45-52

pkg/controller/unifiedpushserver/postgresql.go
grdryn: > I guess this depends how the ops manage this situation. Perhaps the Postgres version they want to use for UPS is different than the version they want to use in some other place.

For our case at least, I believe we'll always want to use an image from the centralized PostgreSQL ImageStream. Different versions can be (and are) specified on that ImageStream with different ImageStreamTags.

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: Huh, that's interesting, it's there when I look on OCP/OSD. I guess it's fine to create one then. Given that that ImageStream would just be used just for UPS, I think it might actually be a good thing for it to have the `importPolicy.scheduled = true`, so that we don't have to instruct operations people that when they `oc import-image` for their other ImageStreams that they need to remember to do this one also.

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: That's interesting! :thinking: 

Does it still work if you set `Automatic: false`? If the trigger is required (which it sounds like it is) then it sounds like in our case setting automatic to be false might help. Unfortunately the API doesn't seem to specify what the default value of that flag is: https://github.com/openshift/origin/blob/8da9ab8f033427d7d463308fc7035e95fd20a7b4/staging/src/github.com/openshift/openshift-apiserver/pkg/apps/apis/apps/types.go#L277

pkg/controller/unifiedpushserver/postgresql.go
aliok: Yeah, using the OpenShift ImageStream in later commits.

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: Ok, so 
OauthProxy --> scheduled=true
UPS --> scheduled=false

Is this what you think?

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: Tried `Automatic=false` already. Doesn't work. It simply doesn't do the deployment of the init proxy in that case (ie. no pods created with a message like "waiting for something to happen".)

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: Ugh, that does sound insane so maybe not. I guess set them both to false based on @maleck13's previous guidance.

@aliok do you have any thoughts on what they should be set to? I have to admit to not really liking the inclusion of imagestreams in the first place, so I'm just trying to make sure we actually satisfy the requirements from those that have foisted it upon us. I guess because UPS is our responsibility though, we have some say in the matter: do we want clusters to update their UPS deployments automatically when there's a new UPS or oauth-proxy image in the registry, or should they only do that in a maintenance window?

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: Hmm, okay, in that case I think we have two options, and either are fine:

- allow it to trigger the UPS deployment when there's a new Postgresql image, and that's fine because the postgresql will be redeployed anyway, so UPS won't be available while that's happening.
- Don't use an ImageStream for this init container: it seems more correct to not trigger for this than to trigger, since in a successfully running pod this image isn't running. If we leave it at a floating tag, with an ImagePullPolicy set to Always, I think we get the behaviour that we want (no unexpected redeploy, use the newest image the next time it's needed).

What do you think?

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: IMO, our SRE's should decide on this. I was trying to get their opinion on similar matters with an email on AeroGear ML but I guess I wasn't successful. Let's do `scheduled=false` for now for both then.

pkg/controller/unifiedpushserver/unifiedpushserver.go
aliok: IMO, option 1 is better. Otherwise there will be more confusion around the Postgres image we use. I personally wouldn't like it when there is an ImageStream we use for Postgres as well as a regular Image.

deploy/role_binding.yaml
aliok: I deleted this one and things continued to work.
Probably a left over when they wanted to test image tagging on OpenShift.
@odra  FYI

aliok:
@grdryn addressed all comments and resolved the conversations.

@psturc can you verify this PR?

psturc:
👀

psturc:
In general LGTM, I verified the functionality.

Just one small thing: I was seeing some (frequently occurring) error in operator's log: https://jsonblob.com/2b775873-9745-11e9-9918-b7c2cf892003

aliok:
@psturc I think that error message was shown for other resources already. I suggest we address that separately.

aliok:
Jenkins job was stuck, restarted it now

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -375,70 +383,83 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+	if err := controllerutil.SetControllerReference(instance, oauthProxyImageStream, r.scheme); err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// Check if this Deployment already exists
-	foundUnifiedpushDeployment := &appsv1.Deployment{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: unifiedpushDeployment.Name, Namespace: unifiedpushDeployment.Namespace}, foundUnifiedpushDeployment)
+	// Check if this ImageStream already exists
+	foundOauthProxyImageStream := &imagev1.ImageStream{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: foundOauthProxyImageStream.Name, Namespace: oauthProxyImageStream.Namespace}, foundOauthProxyImageStream)
```
grdryn, line 401: I think this line here is the cause of the errors in the logs that @psturc noticed. the `NameSpacedName` here should be using `oauthProxyImageStream.Name` rather than `foundOauthProxyImageStream.Name`. I'll fix it in my next PR.

aliok: +1

PR #31, opened 2019-06-26
UPS Universal container and enmasse 

secondsun:
This PR adds support for using AMQ Online as a message broker.  This allows UPS to run in a multitenancy fashion.

# Testing notes
Until we Publish the new UPS container, I have provided a docker formatted container that is based on the latest version of UPS.  To use this image set the ups initial image stream using `export UPS_IMAGE_STREAM_INITIAL_IMAGE=docker.io/secondsun/ups:3.0.0`

#Testing Steps

1. Run the operator with the environment variable `export UPS_IMAGE_STREAM_INITIAL_IMAGE=docker.io/secondsun/ups:3.0.0` set
2. Create a UPS server using the example as described in the readme
3. Verify the server deploys by checking the unifiedpush and AMQ Online projects for the correct resources. 
4. Create a Android or iOS variant using whatever method you feel appropriate
5. Send a message using UPS
6. Verify that the message was sent using AMQ online.  You can do this by looking at the AMQ Online console in the ups address space.  You should see on the addresses for your messaging queues that messages have been sent on those addresses
7. Undeploy UPS
8. verify that the AMQ and UPS  resources cleanup 

# Screen Shot of AMQ console showing FCM message activity 
image


deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -8,8 +8,6 @@ spec:
     kind: UnifiedPushServer
     listKind: UnifiedPushServerList
     plural: unifiedpushservers
-    shortNames:
```
grdryn: Could you check what version of the `operator-sdk` command you've got? This shouldn't get removed, but I can't remember if the fix is in a released version or not :-/

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -69,6 +67,9 @@ spec:
                 - backendSecretName
                 type: object
               type: array
+            useMessageBroker:
+              description: 'Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html'
```
grdryn: This description gets generated from the comment on the field in `pkg/apis/push/v1alpha1/unifiedpushserver_types.go`, so please either change or remove that.

deploy/ups-infra/ups-queue.yaml
```diff
@@ -0,0 +1,2 @@
+
```
grdryn: This file looks odd, is it intentionally empty?

deploy/ups-infra/users.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: user.enmasse.io/v1beta1
+kind: MessagingUser
+metadata:
+  name: ups.upsuser
+spec:
+  username: upsuser
+  authentication:
+    type: password
+    password: cGFzc3dvcmQ= # Base64 encoded
```
grdryn: Is this intentionally hard-coded, or should it be generated?

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
```diff
@@ -14,6 +14,9 @@ type UnifiedPushServerSpec struct {
 
 	// Backups is an array of configs that will be used to create CronJob resource instances
 	Backups []UnifiedPushServerBackup `json:"backups,omitempty"`
+
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
```
grdryn: This is the comment that I mentioned earlier, that should be either changed or removed (it's added as the description of the field in the CRD.

Makefile
```diff
@@ -19,6 +19,12 @@ TEST_COMPILE_OUTPUT = build/_output/bin/unifiedpush-operator-test
 code/run: code/gen
 	operator-sdk up local
 
+.PHONY: code/debug
+code/debug: code/gen
+	export WATCH_NAMESPACE=unifiedpush
+	dlv --listen=:2345 --headless=true --api-version=2 exec ./build/_output/bin/unifiedpush-operator-local  --
```
grdryn: I don't expect there's enough in the pre-requisite `code/gen` target to make `./build/_output/bin/unifiedpush-operator-local`, is there? :thinking: If not, then you could change the pre-requisite target on L23 to be `code/compile`, and change the binary here to be `./build/_output/bin/unifiedpush-operator`.

You'll be happy to know that the next version of `operator-sdk` will have `delve` support built-in, with `operator-sdk up local --enable-delve`, so you won't need to remember to build a new version of the binary beforehand like this. :smiley_cat:

Gopkg.toml
```diff
@@ -73,3 +73,8 @@ required = [
   [[prune.project]]
     name = "k8s.io/gengo"
     non-go = false
+
+[[constraint]]
+  name = "github.com/enmasseproject/enmasse"
+  source = "https://github.com/secondsun/enmasse.git"
+  branch = "address_type_update"
```
grdryn: Is there a PR to get your changes back to the enmasse project? If so, could you link it here? If not, should we maybe change so that we're referencing your fork explicitly in imports?

pkg/controller/unifiedpushserver/enmasse.go
```diff
@@ -0,0 +1,92 @@
...
+		Spec: enmassev1beta.AddressSpec{
+			Address: address,
+			Type:    "topic",
+			Plan:    "brokered-topic",
+		},
+	}
+}
+
+func newMessagingUser(cr *pushv1alpha1.UnifiedPushServer) *messaginguserv1beta.MessagingUser {
+	password := []byte("password")
```
grdryn: Like in the equivalent yaml file, should this be generated? There's a basic `generatePassword` function here that might help: https://github.com/aerogear/unifiedpush-operator/blob/master/pkg/controller/unifiedpushserver/util.go#L29

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -89,7 +89,85 @@ func newOauthProxyImageStream(cr *pushv1alpha1.UnifiedPushServer) (*imagev1.Imag
...
+						Name: fmt.Sprintf("%s-postgresql", cr.Name),
+					},
+				},
+			},
+		},
+	}
+
+	if cr.Spec.UseMessageBroker {
+		env = append(env,
+			corev1.EnvVar{
```
grdryn, line 146: When the `MessagingUser` is created, are these values somehow put into a secret that we could determine the name of? If so, maybe we should reference them directly from the secret fields, like we do for the postgresql env vars above.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -118,6 +119,32 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+	// log.Info("Registering AMQ watches address")
```
grdryn: TODO: something about this commented code.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		if requeueCreate {
+			return reconcile.Result{Requeue: true}, nil
+		} else {
+			reqLogger.Info("Found All queues and topics for UPS")
+		}
+
+	}
+	//If AMQ is used, it is ready
 
 	//#region Postgres PVC
```
grdryn, line 377: Check out these cool separators that @aliok found: apparently some of the major editors/IDEs allow you to fold these regions.

It's really just a hack until we bother to find a good abstraction for this stuff, but it might be good to wrap the stuff in this PR in a region block.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -445,7 +594,7 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 	//#endregion
 
 	//#region UPS DeploymentConfig
-	unifiedpushDeploymentConfig, err := newUnifiedPushServerDeploymentConfig(instance)
+	unifiedpushDeploymentConfig, err := newUnifiedPushServerDeployment(instance, addressSpaceURL)
```
grdryn: We previously had the function name as `newUnifiedPushServerDeployment`, but Ali changed it when adding the ImageStream support (it changed from creating a K8s Deployment to creating an OpenShift DeploymentConfig). It doesn't really matter which one it's called, just highlighting it!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 			return reconcile.Result{}, err
 		}
 	}
+	addressSpaceURL := ""
+	//Begin AMQ resource reconcile
+	if instance.Spec.UseMessageBroker {
+		//check that address space exists
+		foundAddressSpace := newAddressSpace(instance)
```
```suggestion
		addressSpace := newAddressSpace(instance)
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+
+		// Set UnifiedPushServer instance as the owner and controller
+		if err := controllerutil.SetControllerReference(instance, foundAddressSpace, r.scheme); err != nil {
```
```suggestion
		if err := controllerutil.SetControllerReference(instance, addressSpace, r.scheme); err != nil {
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+			return reconcile.Result{}, err
+		}
+
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: foundAddressSpace.Name, Namespace: foundAddressSpace.Namespace}, foundAddressSpace)
```
```suggestion
		foundAddressSpace := &enmassev1beta.AddressSpace{}
		err = r.client.Get(context.TODO(), types.NamespacedName{Name: addressSpace.Name, Namespace: addressSpace.Namespace}, foundAddressSpace)
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Creating a new Address Space", "AddressSpace.Namespace", foundAddressSpace.Namespace, "AddressSpace.Name", foundAddressSpace.Name)
+			err = r.client.Create(context.TODO(), foundAddressSpace)
```
```suggestion
			err = r.client.Create(context.TODO(), addressSpace)
```

This, and the other suggested comments above it, are intended to make it more like the other sections here. It probably works fine this way, but I'm not sure if the call to `r.client.Get` would ever clobber the `foundAddressSpace` when trying to read into it, in a way that could break this call to `r.client.Create`.

If this is considered a better way, then I think the name would be better without the `found` prefix, and we should consider the same pattern for the other creations that we do here also.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{Requeue: true}, nil
+		} else if err != nil {
+			return reconcile.Result{}, err
+		} else if !foundAddressSpace.Status.IsReady {
+			return reconcile.Result{Requeue: true}, nil
```
grdryn: Do you have any indication of how long this could take? If it's a while, maybe a `RequeueAfter: time.Second * N` would be better than a straight up `Requeue: true`, even if your value of N is as low as 1? It probably doesn't make too much of a difference either way.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 			return reconcile.Result{}, err
 		}
 	}
+	addressSpaceURL := ""
```
grdryn: Consider moving this declaration to right before the attempt to assign a value to it in the range on line 255

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -196,10 +223,133 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		} else if !foundAddressSpace.Status.IsReady {
+			return reconcile.Result{Requeue: true}, nil
+		} else {
+			reqLogger.Info("Found AddressSpace for UPS")
+		}
+
+		/*Address space exists and is ready*/
+
+		for _, status := range foundAddressSpace.Status.EndpointStatus {
+			if status.Name == "messaging" { //magic value
```
grdryn: Could we ever get to the end of this range loop without having set a value, or will there always be an EndpointStatus called messaging?

Gopkg.toml
```diff
@@ -73,3 +73,8 @@ required = [
   [[prune.project]]
     name = "k8s.io/gengo"
     non-go = false
+
+[[constraint]]
```
grdryn: For neatness, I'd suggest adding this above the `[prune]` section. Additionally, if there are a bunch of packages in it that we don't make use of, you could add a `[[prune.project]]` section under `[prune]` that will limit the number of useless files that we check in (I wish I'd known about this before checking in the vendor dir initially...for the already committed stuff it wouldn't help that much to remove them now):

```toml
  [[prune.project]]
    name = "github.com/enmasseproject/enmasse"
    unused-packages = true
```

grdryn (commented):
I didn't get to review this entire thing (I didn't get all the way down the Reconcile method), and I haven't verified it yet (I'll hopefully get a chance to do that in my morning before you start).

There are a bunch of yaml files that are added here in `deploy/ups-infra`, but I don't think they're used, are they? The equivalent resources appear to be generated now by the controller (this is better). Are the yaml files still needed?

grdryn:
@odra Do we need to consider this in the context of backups? Should backup of stuff in enmasse for UPS be dealt with in our CR like the PostgreSQL backup? Should it be part of the same backup entry or a new one (`spec.backups` in the CR is a list)?

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
secondsun: operator-sdk version: v0.7.0
Should I build operator-sdk from master?

deploy/ups-infra/users.yaml
secondsun: IT can be generated.  These files aren't used but it is hardcoded in the operator; I will change that.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: address space is usually pretty slow (a minute or two) I'll give it 60 seconds.

odra:
There is an enmasse backup job already which archives everything in the enmasse namespace, is this ups enmasse stuff created in a different ns?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: I'll need to pair on this one; the address space url is conditionally used when we make the deployment object and will be needed then.  I suppose I can use some form of selector or something similar to the secret selectors.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: There will always be one called messaging, see https://enmasse.io/documentation/0.28.2/openshift/

secondsun:
@odra This creates messaginguser, addressspace, and address resources in the ups namespace.  Those resources are consumed by services running in the enmasse namespace if that makes sense.

odra:
So all data as broker pods, queues, etc are already backed up by the enmasse cronjob - restoring enmasse would restore it for ups as well

deploy/ups-infra/ups-queue.yaml
secondsun: File was acidentally kept in after the rebase, sorry!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: I don't remember why I put it in there; I'll take it out.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: I don't really care either way, it was just a problem when I was rebasing.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: Let me make sure I'm following you;

1. addressSpace is the object we want to have
2. foundAddressSpace is an object that possibly exists inside of k8s 
2.a If there is a IsNotFound error getting foundAddressSpace then use addressSpace to create a new addressSpace
2.b otherwise check the status of foundAddressSpace and continue or requeue as necessary.

I see the problems you are talking about with the names being shit ;)

Makefile
camilamacedo86: ### The Env var probably will be not added after it. 
- Run the command then do an echo on the ENV VAR.
- See here in MSS OPER how we solve it. https://github.com/aerogear/mobile-security-service-operator/blob/master/Makefile#L169

### The debug config can change according to IDE. 
Also, I'd not recommend add the dvl here. I am not sure it will work for all IDE, so see how we did it in MSS Oper here https://github.com/aerogear/mobile-security-service-operator#debugging. 

### It is missing the ENV VARs 
- Where the app can be applied
- Operator namespace

c/c @grdryn @secondsun

secondsun: It works; I've been using it.  The mechanism is using a socket to attach a debugger, it works in all IDEs much more so than anything else I've tried.

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
camilamacedo86: The latest version is 0.8.1. What version is used in the project? Could we use here 0.8.1?

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
camilamacedo86: Yep. What you add on top of the attribute of the CR/CRD will be added automatically here.

secondsun:
@grdryn @camilamacedo86 I've updated and rebased care to check again?

camilamacedo86:
@damienomurchu could you give a hand here?

Why I do the PR the CI and coveralls are called as here <https://github.com/aerogear/unifiedpush-operator/pull/47> for example and when others do it not? Could you help me let me know what is not setup?

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
```diff
@@ -14,6 +14,9 @@ type UnifiedPushServerSpec struct {
 
 	// Backups is an array of configs that will be used to create CronJob resource instances
 	Backups []UnifiedPushServerBackup `json:"backups,omitempty"`
+
+	// If you are using enmasse you can set this value to true to use managed queues.
```
grdryn:
```suggestion
	// UseMessageBroker can be set to true to use managed queues, if you are using enmasse. Defaults to false.
```

Generally Go doc comments should start with the name of the thing being documented for searchability (and generally "convention). Additionally, when you run `make code/gen`, it will update the descriptoin in the crd yaml.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+
+		foundAddressSpace := &enmassev1beta.AddressSpace{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: addressSpace.Name, Namespace: addressSpace.Namespace}, foundAddressSpace)
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Creating a new Address Space", "AddressSpace.Namespace", addressSpace.Namespace, "AddressSpace.Name", addressSpace.Name)
+			err = r.client.Create(context.TODO(), addressSpace)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			return reconcile.Result{RequeueAfter: time.Second * 10}, nil
```
grdryn, line 249: I'm guessing that this is needed so that we come back in 10 seconds to see if it's ready? If so, could you add a log message above it there as some sort of documentation (so that future me or you knows and isn't worried about it just being an assumption).

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		} else if err != nil {
+			return reconcile.Result{}, err
+		} else if !foundAddressSpace.Status.IsReady {
+			return reconcile.Result{RequeueAfter: time.Second * 10}, nil
```
grdryn, line 254: A similar log message could be added above this return.

pkg/controller/unifiedpushserver/unifiedpushserver.go
grdryn: I see that now there's a Secret (for the password) and a ConfigMap (for the host). Are those just being created so that they can be referenced in these EnvVars in the UPS deploymentconfig? If so, that may or may not be worthwhile. :thinking: What I was wondering in the question above was more like "incidentally, if they happen to be in a Secret or ConfigMap, we could wire them up from there".

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+				if err != nil && !errors.IsAlreadyExists(err) {
+					return reconcile.Result{}, err
+				}
+
+			}
+		}
+
+		//check that user exists
+		user, err := newMessagingUser(instance)
+
```
```suggestion
```

Visually we tend to have the `if err != nil {` directly on the next line, rather than having a blank line. This will reduce the chances that someone will assume they can just put something between the function call and the error check, rather than after the error check.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		foundUser := &messaginguserv1beta.MessagingUser{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: user.Name, Namespace: user.Namespace}, foundUser)
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Creating a new MessagingUser", "MessagingUser.Namespace", user.Namespace, "MessagingUser.Name", user.Name)
+			err = r.client.Create(context.TODO(), user)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			secret := newAMQSecret(instance, string(user.Spec.Authentication.Password))
```
grdryn: I'd pull out this secret reconciliation into its own section (say, after line 274), and do the usual rigamarole of checking if it exists first. `user` should still be in scope there.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+			err = r.client.Create(context.TODO(), secret)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+
+			return reconcile.Result{RequeueAfter: time.Second * 1}, nil
```
grdryn: At this point, the MessagingUser (and Secret) have been successfully created. Is it necessary to return-and-requeue here, or should it continue straight away to the queues stuff? If a return-and-requeue is needed, is it necessary to wait a second (rather than `RequeueAfter: time.Second * 1`, you could have `Requeue: true`).

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		} /*User exists and is ready*/
+
+		/*User exists and is ready*/
+		//check that addresses exist
+		//queues
+		queues := []string{"APNsPushMessageQueue", "APNsTokenBatchQueue", "GCMPushMessageQueue", "GCMTokenBatchQueue", "WNSPushMessageQueue", "WNSTokenBatchQueue", "MetricsQueue", "TriggerMetricCollectionQueue", "TriggerVariantMetricCollectionQueue", "BatchLoadedQueue", "AllBatchesLoadedQueue", "FreeServiceSlotQueue"}
+		requeueCreate := false
+		for _, address := range queues {
+			queue := newQueue(instance, address)
+			foundQueue := &enmassev1beta.Address{}
```
grdryn, line 312: I'm confused between "queue" and "topic" and "address". Are these all synonyms in enmasse? From my reading here, I'm guessing that "queue" and "topic" are types of "address". Is that right?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+				reqLogger.Info("Queue Not ready", "Queue.Name", foundQueue.Name)
+				requeueCreate = true
+			}
+		}
+
+		if requeueCreate {
+			return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+		}
+
+		reqLogger.Info("Found All queues  for UPS")
```
```suggestion
		reqLogger.Info("Found all queues for UPS")
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+
+		if requeueCreate {
+			return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+		}
+
+		reqLogger.Info("Found All queues and topics for UPS")
+
+	}
+	//If AMQ is used, it is ready
+
```
grdryn, line 376:
```suggestion
	//#endregion

```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 		}
 	}
 
+	//Begin AMQ resource reconcile
```
```suggestion
	//#region AMQ resource reconcile
```

This, in conjunction with the related `//#endregion` after this block, will allow certain IDEs to visually fold this block.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,9 +200,151 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+
+		reqLogger.Info("Found All queues and topics for UPS")
+
+	}
+	//If AMQ is used, it is ready
+
 	//#region Postgres PVC
 	persistentVolumeClaim, err := newPostgresqlPersistentVolumeClaim(instance)
 	if err != nil {
+		reqLogger.Info("Error making PVC")
```
grdryn: This seems unrelated to everything else. Have you had issues with it (and the logs that would be caused by returning with the error in the line below)? The log message, without displaying anything about the error, seems incomplete to me.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -445,7 +589,7 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 	//#endregion
 
 	//#region UPS DeploymentConfig
-	unifiedpushDeploymentConfig, err := newUnifiedPushServerDeploymentConfig(instance)
+	unifiedpushDeploymentConfig, err := newUnifiedPushServerDeployment(instance)
```
grdryn: I believe you said this function name difference was as a result of a rebase conflict. Should it be changed back?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: They are both addresses with slightly different creation semantics that I wasn't comfortable abstracting out.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: I combined the configmap and secret into just a secret and moved it to its own section

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -197,6 +200,153 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		//#region create addressSpace
+		addressSpace := newAddressSpace(instance)
+
+		// Set UnifiedPushServer instance as the owner and controller
+		if err := controllerutil.SetControllerReference(instance, addressSpace, r.scheme); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		foundAddressSpace := &enmassev1beta.AddressSpace{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: addressSpace.Name, Namespace: addressSpace.Namespace}, foundAddressSpace)
```
grdryn, line 241: Some additional permissions will need to be granted in `deploy/role.yaml`. I just tried by deploying into a cluster, which uses the ServiceAccount (rather than whatever admin user you are logged in as when running locally), and got this persistent error:

```
E0717 10:50:51.724744       1 reflector.go:134] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:196: Failed to list *v1beta1.AddressSpace: addressspaces.enmasse.io is forbidden: User "system:serviceaccount:unifiedpush:unifiedpush-operator" cannot list addressspaces.enmasse.io in the namespace "unifiedpush": no RBAC policy matched
```

This also reminds me that these new types like `AddressSpace`, etc, are now secondary resource types for this controller, and should have watches set in the `Add` function above.

grdryn (changes requested):
When I run the operator locally against an integreatly installation, there's a lot of logs like this:

```
E0717 12:23:48.823452   31518 streamwatcher.go:109] Unable to decode an event from the watch stream: unable to decode to metav1.Event
```

This is not normal in my experience. Have you seen it before? Maybe it will go away when we add the missing watches, but I'm not sure.

I've also added an inline comment about the missing watches. I'm happy to :pear: later if you want help to get this over the line :+1: 

Edit: it also looks like this may need to be rebased again.

grdryn (approved):
I have reviewed and verified this.

I'll see if I can determine why the Jenkins build is failing. Once we get that to pass, this can be merged.

grdryn:
@psturc any idea what's going wrong here? I see that you've made changes to the library step that's currently failing (https://github.com/aerogear/qe-pipeline-library/blob/master/vars/runOperatorTestWithImage.groovy), but it seems to work fine on master and fail here.

The error is below. I'm not sure if that message about not being able to build the test binary, since there's an earlier stage that calls `make test/compile` which should do that. Wdyt?

<details>
+ operator-sdk test local ./test/e2e --namespace unifiedpush --go-test-flags -v
time="2019-07-17T12:06:59-04:00" level=info msg="Testing operator locally."
=== RUN   TestUnifiedpush
=== RUN   TestUnifiedpush/unifiedpush-e2e
--- FAIL: TestUnifiedpush (0.10s)
    --- FAIL: TestUnifiedpush/unifiedpush-e2e (0.04s)
        client.go:57: resource type ServiceAccount with namespace/name (unifiedpush/unifiedpush-operator) created
        client.go:57: resource type ClusterRole with namespace/name (unifiedpush/unifiedpush-operator) created
        unifiedpush_test.go:40: failed to initialize cluster resources: the server could not find the requested resource
        client.go:75: resource type ClusterRole with namespace/name (unifiedpush/unifiedpush-operator) successfully deleted
        client.go:75: resource type ServiceAccount with namespace/name (unifiedpush/unifiedpush-operator) successfully deleted
FAIL
FAIL	github.com/aerogear/unifiedpush-operator/test/e2e	0.253s
Error: failed to build test binary: (failed to exec []string{"go", "test", "./test/e2e/...", "-namespacedMan", "/tmp/namespaced-manifest.yaml088960031", "-globalMan", "/tmp/global-manifest.yaml456540914", "-root", "/home/jenkins/workspace/unifiedpush-operator_PR-31-SEY6RPEIZWBCL3AAUZIKQLVIRCT3ZHUB2I4EFPDJLHEWTTHLMGWQ/src/github.com/aerogear/unifiedpush-operator", "-v", "-singleNamespace", "-parallel=1"}: exit status 1)
Usage:
  operator-sdk test local <path to tests directory> [flags]

Flags:
      --debug                        Enable debug-level logging
      --global-manifest string       Path to manifest for Global resources (e.g. CRD manifests)
      --go-test-flags string         Additional flags to pass to go test
  -h, --help                         help for local
      --image string                 Use a different operator image from the one specified in the namespaced manifest
      --kubeconfig string            Kubeconfig path
      --molecule-test-flags string   Additional flags to pass to molecule test
      --namespace string             If non-empty, single namespace to run tests in
      --namespaced-manifest string   Path to manifest for per-test, namespaced resources (e.g. RBAC and Operator manifest)
      --no-setup                     Disable test resource creation
      --up-local                     Enable running operator locally with go run instead of as an image in the cluster

Global Flags:
      --verbose   Enable verbose logging
</details>

grdryn:
The issue was the new structure of the role.yaml. I fixed it and pushed a fix here, and the build passed so I merged.

secondsun:
@grdryn You are a gentleman and a scholar!

PR #32, opened 2019-06-26
Optionally create backup CronJob(s) based on UPS CR

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9433

----

# Verification steps

* Check out this PR
* Deploy the operator
```
make cluster/prepare
cat deploy/operator.yaml | sed -e "s|quay.io/aerogear/unifiedpush-operator:master|quay.io/aerogear/unifiedpush-operator:$(git rev-parse --short HEAD)|g" | kubectl create -n unifiedpush -f -
```
* Create a ServiceAccount called `backupjob`. This is needed for the resulting cronjobs, but is not created by the operator, since it may need additional access that the operator may not be able to give -- if the secrets being used are in different namespaces, then there would need to be relevant rolebindings for it, so we deliberately leave out the creation of all of that stuff:
```
kubectl create sa backupjob
```
* Create a UPS CR that has backups specified:
```
kubectl apply -f deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```

* Verify that there are no persistent errors in the operator logs, and that everything gets created as expected. The important things to check for here are "cronjob" objects:

```
kubectl get cronjobs -l app=example-ups-with-backups
NAME                    SCHEDULE    SUSPEND   ACTIVE    LAST SCHEDULE   AGE
ups-daily-at-midnight   0 0 * * *   False     0         <none>          <invalid>
```

* Edit the schedule of the cronjob, and notice that it gets changed back immediately (note the output may not look exactly like this, due to concurrency, I've annotated with comments):
```
$ oc get cronjobs -l app=example-ups-with-backups -w & oc patch cronjob ups-daily-at-midnight -p '{"spec": {"schedule": "1 0 * * *"}}'
[1] 17231 #pid of backgrounded `oc get -w` task, you can kill it after this step, e.g. `kill 17231`
NAME                    SCHEDULE    SUSPEND   ACTIVE    LAST SCHEDULE   AGE
ups-daily-at-midnight   0 0 * * *   False     0         <none>          4m #before patch, as specified in CR
cronjob.batch/ups-daily-at-midnight patched
ups-daily-at-midnight   1 0 * * *   False     0         <none>    4m #after patch
ups-daily-at-midnight   0 0 * * *   False     0         <none>    4m #after reconcile fixes it back to what the CR specified
```

* Update the schedule for a backup in the CR, and notice that the cronjob gets updated immediately, as expected (just use `kubectl edit ups example-ups-with-backups` for this...I didn't spend long enough to figure out the `kubectl patch` for it).

@odra has already verified that the backups work as expected. If you want to do that also, you'll need to configure secrets (the one for AWS details at least, and the one for PGP if you want to verify that). It's not worth the effort, in my opinion, given that Leo has already done it (thanks Leo!).


README.adoc
```diff
@@ -100,8 +100,24 @@ This operator currently only supports one UnifiedPushServer CR to be
 created.
 ====
 
-There are currently no configurable fields in the UnifiedPushServer CR
-apart from the name, so the example in
+Here are all of the configurable fields in a UnifiedPushServer:
```
grdryn, line 103: image
^Click the rich diff button at the top right of this file to see how this will render.

.gitattributes
```diff
@@ -0,0 +1 @@
+**/zz_generated.*.go linguist-generated=true
```
grdryn, line 1: image
Works! :smiley:

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: push.aerogear.org/v1alpha1
+kind: UnifiedPushServer
+metadata:
+  name: example-ups-with-backups
+spec:
+  backups:
```
grdryn, line 6: When thinking about this originally, I was thinking about something more flexible like this: https://gist.github.com/grdryn/034d654a32af7dadd79ba1c47c17ee60

Here in this PR though, I've just copied the integreatly way, as done in the keycloak operator, for consistency. Unfortunately that ties us to some things that are not very transparent (and required me reading most of the code in [integr8ly/backup-container-image](https://github.com/integr8ly/backup-container-image) to see what was needed. Hence the elaborate comments in here.

pkg/config/config.go
```diff
@@ -19,6 +20,8 @@ type Config struct {
 	UPSImageStreamInitialImage        string
 	PostgresImageStreamInitialImage   string
 	OauthProxyImageStreamInitialImage string
+
+	BackupImage string
```
grdryn, line 24: Not using an imagestream for this one, since it's only used in a CronJob rather than a DeploymentConfig.

pkg/config/config.go
```diff
@@ -27,18 +30,21 @@ func New() Config {
...
+		UPSImageStreamName: getEnv("UPS_IMAGE_STREAM_NAME", "ups-imagestream"),
+		UPSImageStreamTag:  getEnv("UPS_IMAGE_STREAM_TAG", "latest"),
+
 		OauthProxyImageStreamName: getEnv("OAUTH_PROXY_IMAGE_STREAM_NAME", "ups-oauth-proxy-imagestream"),
 		OauthProxyImageStreamTag:  getEnv("OAUTH_PROXY_IMAGE_STREAM_TAG", "latest"),
 
 		PostgresImageStreamNamespace: getEnv("POSTGRES_IMAGE_STREAM_NAMESPACE", "openshift"),
 		PostgresImageStreamName:      getEnv("POSTGRES_IMAGE_STREAM_NAME", "postgresql"),
-		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
+		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "9.6"),
```
grdryn: The `latest` ImageStreamTag in that ImageStream provides PostgreSQL 10. Maybe that's fine, but we've only really used 9.6 up to now. @secondsun WDYT?

pkg/controller/unifiedpushserver/postgresql.go
```diff
@@ -41,9 +41,11 @@ func newPostgresqlSecret(cr *pushv1alpha1.UnifiedPushServer) (*corev1.Secret, er
 	return &corev1.Secret{
 		ObjectMeta: objectMeta(cr, "postgresql"),
 		StringData: map[string]string{
-			"database-name":     "unifiedpush",
-			"database-user":     "unifiedpush",
-			"database-password": databasePassword,
+			"POSTGRES_DATABASE":  "unifiedpush",
```
grdryn, line 44: The [integr8ly/backup-container-image](https://github.com/integr8ly/backup-container-image) needs this to be in a specific format. Also, if we do decide to use PostgreSQL 10 (there's a different comment about that in another file in this PR), then we need to specify `POSTGRES_VERSION` in here too.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -398,7 +410,7 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 
 	// Check if this ImageStream already exists
 	foundOauthProxyImageStream := &imagev1.ImageStream{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: foundOauthProxyImageStream.Name, Namespace: oauthProxyImageStream.Namespace}, foundOauthProxyImageStream)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: oauthProxyImageStream.Name, Namespace: oauthProxyImageStream.Namespace}, foundOauthProxyImageStream)
```
grdryn: This is also available as a standalone PR in #33. When that's merged I'll rebase on top of it.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -459,6 +471,41 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 	}
 	//#endregion
 
+	//#region Backups
```
grdryn, line 471: This doesn't yet do anything to reconcile the backups after the first creation, which it really should (e.g. someone changing the schedule for a backup should cause it to update (rather than them having to update the generated CronJob, or worse, having to delete and recreate the UnifiedPushServer CR). Similarly adding/removing backup entries should cause the appropriate CronJobs to be added/removed.

I haven't added that yet, because there's a regression here in our monstrous reconcile function: once the `status.phase` is set to `Complete`, we ignore any other changes (both on the cluster and to the CR). This means that even if someone deleted the UPS or PostgreSQL deployments, we wouldn't be recreating them, which is not what we want. I'll create another issue for that and link it here (and also mention that then updates to backup entries or the resulting CronJobs should be handled appropriately).

pkg/config/config.go
```diff
@@ -27,18 +30,21 @@ func New() Config {
...
 
 		// these are used when the image stream does not exist and created for the first time by the operator
 		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
 		OauthProxyImageStreamInitialImage: getEnv("OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/openshift/oauth-proxy:v1.1.0"),
+
+		BackupImage: getEnv("BACKUP_IMAGE", "quay.io/integreatly/backup-container:master"),
```
odra: are we always going to use master? is there value in using it as a cr param?

pkg/controller/unifiedpushserver/backups.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return cronjobs, nil
+}
+
+func buildBackupContainerCommand(upsBackup pushv1alpha1.UnifiedPushServerBackup, upsNamespace string) []string {
+	command := []string{"/opt/intly/tools/entrypoint.sh", "-c", "postgres", "-n", upsNamespace}
+
+	// If there is no encryption secret, we need to inhibit the
+	// encryption behaviour
+	if upsBackup.EncryptionKeySecretName == "" {
```
odra, line 63: does the operator deals with a situation where this var is filled?

pkg/controller/unifiedpushserver/backups.go
```diff
@@ -0,0 +1,130 @@
...
+func buildBackupCronJobEnvVars(upsBackup pushv1alpha1.UnifiedPushServerBackup, upsName string, upsNamespace string) []corev1.EnvVar {
+
+	envVars := []corev1.EnvVar{
+		{
+			Name:  "PRODUCT_NAME",
+			Value: "unifiedpush",
+		},
+		{
+			Name:  "COMPONENT_SECRET_NAME",
+			Value: fmt.Sprintf("%s-%s", upsName, "postgresql"),
```
odra, line 79: can you use it as `"%s-%s-secret"` as it is used in other products?

pkg/config/config.go
grdryn: I'm not sure what tag we should use by default here, maybe a floating `1.0` tag that moves with patch releases to that stream would be better than master (which could break things here)? Having it at specific releases could fall out-of-date quickly.

I don't think it should be part of the CR though. It's configurable through the `BACKUP_IMAGE` env var to the operator here.

pkg/controller/unifiedpushserver/backups.go
grdryn: By default that `-e` option to the image gets a value of `gpg`: https://github.com/integr8ly/backup-container-image/blob/master/image/tools/entrypoint.sh#L7

That means that if we don't specify any other value, it will try to encrypt the backups. We have to explicitly set it to an empty string here to get it to not try encryption (the backup job would fail otherwise).

pkg/controller/unifiedpushserver/backups.go
grdryn: We could, but it seems redundant to me to add the kind to the name :thinking: Is there a benefit to doing that?

Do we add `-deploymentconfig` to the end of DeploymentConfig names? Would you add `-string` to the end of a String value in a typed programming language?

pkg/config/config.go
odra: I am asking because that image changes quite often from a release perspective and how easy would it be to use and maintain  a fixed number rather than use master which may be unstable

pkg/controller/unifiedpushserver/backups.go
odra: yeah but would it be easier to just add it into the above line?

pkg/controller/unifiedpushserver/backups.go
odra: I agree with you but that's how other component secrets are being created today (for now at least)

pkg/config/config.go
grdryn: Yeah, I think you're right. I'll change the default to a fixed tag :+1:

pkg/controller/unifiedpushserver/backups.go
grdryn: I'm not sure what you're suggesting here. Is it to always put the `"-e", ""` in to the `command` slice on line 65? The problem with that is that it means encryption is off, no matter what you specify in the CR (I think the keycloak operator has this problem).

I guess if you want it to be explicitly part of the command string, we could modify this here to be:

```go
if upsBackup.EncryptionKeySecretName == "" {
	command = append(command, "-e", "")
} else {
	command = append(command, "-e", "gpg")
}
```

WDYT?

Unless you're talking about something completely different, which is entirely possible! :see_no_evil:

pkg/controller/unifiedpushserver/backups.go
grdryn: That's good to know. Do you think it will cause us any problems if we don't change?

pkg/controller/unifiedpushserver/backups.go
odra: Sorry I was talking about adding it in this line: https://github.com/aerogear/unifiedpush-operator/pull/32/files#diff-0ccc80f83580c98ef42e35739cdaaef1R65 intead of doing a `if` statement since we don't deal with gpg atm such as: `command := []string{"/opt/intly/tools/entrypoint.sh", "-c", "postgres",  "-e", "", "-n", upsNamespace}`

grdryn: Ah yeah, that's what I thought you meant. If we did that, we should also remove the encryption secret details from the CRD, since it would be misleading (like it is in the keycloak operator).

When the functionality is there in the backup-container-image project, I don't see a problem with having it possible to enable (which adding it unconditionally to the command string like that would mean).

With the way it is here, if someone has correctly specified an encryption secret, then it will work as expected and encrypt. Equally, if they don't specify one, it will work as expected, just not encrypt.

odra: ok makes sense :+1:

pkg/config/config.go
secondsun: It won't be a huge pain in the ass to update, everyone is good about being forward/backwards compatible, and we use Hibernate to handle SQL so in theory there shouldn't be any breaks.

odra:
You need to add the following rules into the role.yaml file:

```
- apiGroups:
    - batch
  resources:
    - "cronjobs"
  verbs:
    - "*"
```

odra:
Postgres data was backed up with success, so I deleted everything and deployed a new ups instance with no data at all.

Postgres data restoration process:

- copy the archive into the pod: `oc cp ~/Downloads/unifiedpush.unifiedpush-07_10_12.pg_dump.gz example-unifiedpushserver-postgresql-1-mfpmj:/tmp`
- rsh into the pod: `oc rsh example-unifiedpushserver-postgresql-1-mfpmj`
- gunzip it: `gunzip unifiedpush.unifiedpush-07_10_12.pg_dump`
- import the archive: `psql -d $POSTGRESQL_DATABASE -f unifiedpush.unifiedpush-07_10_12.pg_dump`
- Restart ups server byt redeploying it

Additional notes:

- Some CRs were not restored since I backed up stuff directly into Postgres, such as a push application and a push variant - should the operator watch for those in ups and create those custom resources?
- I tried deleting a push application cr but got a 404 error (not releated to this pr): https://gist.github.com/odra/78f3286ae9e80b0db8fd06a57da62cb1

pkg/controller/unifiedpushserver/backups.go
odra: not at all, "just a convention", lets keep as it is

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: You're right.

I added that check during my work about limiting the number of UPS servers to 1. But I see that this check I added isn't really required for that. Feel free to remove it.

grdryn:
Thanks for the review & thorough verification @odra! I'll add that missing stuff to the role. I've also added another commit here to make sure we're always generating the openapi stuff also.

I haven't made a commit to the backup-container-image project yet to include the CR kinds from this operator. Is my understanding correct that all CRs would then be backed up by a centralized backup job, and not the database one that we've got here?

Also, at what point in the process did you try deleting a PushApplication CR?

odra:
You are correct, you don't need to run the resources backup here, the backup pod will run it against all namespaces (so all you need is to add ups crs in the backup container script).

I tried deleting it after I crated one pushapplication CR and could see the same app in ups console.

pkg/config/config.go
grdryn: Ok cool, I'll change this to `10` for now then (currently equivalent to latest, but sounds way safer since who knows what an `11` would do).

odra: ups is using 9.6 psql isn't it?

camilamacedo86: I would prefer the image tag as well. It can break with an upper version and with the version configurable by the CR in order to allow more than one product used it and each one with your own version if required.

pkg/config/config.go
```diff
@@ -27,18 +30,21 @@ func New() Config {
...
 		PostgresImageStreamNamespace: getEnv("POSTGRES_IMAGE_STREAM_NAMESPACE", "openshift"),
 		PostgresImageStreamName:      getEnv("POSTGRES_IMAGE_STREAM_NAME", "postgresql"),
-		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
+		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "9.6"),
 
 		// these are used when the image stream does not exist and created for the first time by the operator
 		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
 		OauthProxyImageStreamInitialImage: getEnv("OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/openshift/oauth-proxy:v1.1.0"),
+
+		BackupImage: getEnv("BACKUP_IMAGE", "quay.io/integreatly/backup-container:1.0.8"),
```
camilamacedo86, line 48: Here should be the latest as it is done in other places.

pkg/config/config.go
```diff
@@ -27,18 +30,21 @@ func New() Config {
...
 		OauthProxyImageStreamName: getEnv("OAUTH_PROXY_IMAGE_STREAM_NAME", "ups-oauth-proxy-imagestream"),
 		OauthProxyImageStreamTag:  getEnv("OAUTH_PROXY_IMAGE_STREAM_TAG", "latest"),
 
 		PostgresImageStreamNamespace: getEnv("POSTGRES_IMAGE_STREAM_NAMESPACE", "openshift"),
 		PostgresImageStreamName:      getEnv("POSTGRES_IMAGE_STREAM_NAME", "postgresql"),
-		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "latest"),
+		PostgresImageStreamTag:       getEnv("POSTGRES_IMAGE_STREAM_TAG", "9.6"),
 
 		// these are used when the image stream does not exist and created for the first time by the operator
 		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
```
camilamacedo86, line 45: I'd prefer the implementation encapsulated e the usage specs NOT ENV VARs since it can be optional and the ENV VARs are for the operator and not the backup piece.

pkg/controller/unifiedpushserver/backups.go
```diff
@@ -0,0 +1,130 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func backups(ups *pushv1alpha1.UnifiedPushServer) ([]batchv1beta1.CronJob, error) {
+	cronjobs := []batchv1beta1.CronJob{}
+	for _, upsBackup := range ups.Spec.Backups {
+		cronJobLabels := labels(ups, "backup")
+		for k, v := range upsBackup.Labels {
+			cronJobLabels[k] = v
```
camilamacedo86: Why a loop is required here? Suggestion 

```go
func getBkpLabels(name string) map[string]string { // instance.name
	return map[string]string{"app": "mobilesecurityservice", "mobilesecurityservicebackup_cr": name, "name": "mobilesecurityservicebackup"}
}
```

pkg/controller/unifiedpushserver/backups.go
```diff
@@ -0,0 +1,130 @@
...
+							ObjectMeta: metav1.ObjectMeta{
+								Labels: jobLabels,
+							},
+							Spec: corev1.PodSpec{
+								// This SA needs to be created beforehand
+								// https://github.com/integr8ly/backup-container-image/tree/master/templates/openshift/rbac
+								ServiceAccountName: "backupjob",
+								Containers: []corev1.Container{
+									{
+										Name:            upsBackup.Name + "-keycloak-backup",
```
camilamacedo86: Keylock?

pkg/controller/unifiedpushserver/backups.go
```diff
@@ -0,0 +1,130 @@
...
+				JobTemplate: batchv1beta1.JobTemplateSpec{
+					Spec: batchv1.JobSpec{
+						Template: corev1.PodTemplateSpec{
+							ObjectMeta: metav1.ObjectMeta{
+								Labels: jobLabels,
+							},
+							Spec: corev1.PodSpec{
+								// This SA needs to be created beforehand
+								// https://github.com/integr8ly/backup-container-image/tree/master/templates/openshift/rbac
+								ServiceAccountName: "backupjob",
```
camilamacedo86, line 37: Could you not use the UPS Oper service account?

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+    # Note that backups here is a list, you can specify more than one
+    -
+      # REQUIRED: The name of a backup here will be the name of the
+      # generated CronJob
+      name: ups-daily-at-midnight
```
camilamacedo86, line 11: why the name is required here? Why it is not generated programmatically? (e.g you can use the intance.Name)

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      # OPTIONAL: These labels will be added to the labes of the
+      # CronJob (it will have other standard labels too)
+      labels:
+        "monitoring-key": middleware
```
camilamacedo86: The labels are not required. It is just for ns and should be applied manually by the community or by inegr8ly ansible scripts.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+      schedule: 0 0 * * *
+
+      # These two properties point to a Secret (in this namespace or
+      # somewhere else), that contains data that the scripts in the
+      # backup container can use to upload to a storage location.  The
+      # backup-container-image repo can be found at
+      # https://github.com/integr8ly/backup-container-image and it
+      # currently only supports AWS S3 as a storage backend.
+      #
+      # The Secret here must contain the following keys:
```
camilamacedo86, line 24: If you want you can check the PR: https://github.com/aerogear/mobile-security-service-operator/pull/125.
There have 2 options. If the data be informed the Operator will create the secret if not them will use the applied already in the cluster. (just copy and past should work)

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+      # AWS_S3_BUCKET_NAME
+      # AWS_ACCESS_KEY_ID
+      # AWS_SECRET_ACCESS_KEY
+      #
+      # REQUIRED: Job will fail if not provided
+      backendSecretName: example-aws-key
+      # OPTIONAL: This defaults to the namespace where this CR exists
+      backendSecretNamespace: unifiedpush
```
camilamacedo86: the namespace is not required. It can be obtained programmatically by instance. Namespace.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      # These two properties point to a Secret (in this namespace or
+      # somewhere else), that contains data that the scripts in the
+      # backup container can use to PGP/GPG encrypt the backups before
+      # sending them to the storage location. The
+      # backup-container-image repo can be found at
+      # https://github.com/integr8ly/backup-container-image.
+      #
+      # The Secret here must contain the following keys:
```
camilamacedo86: The same here. https://github.com/aerogear/mobile-security-service-operator/pull/125

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+      # GPG_PUBLIC_KEY
+      # GPG_TRUST_MODEL
+      # GPG_RECIPIENT
+      #
+      # OPTIONAL: If no encryptionKeySecretName is specified, no
+      # encryption attempt will be performed
+      encryptionKeySecretName: example-encryption-key
+      # OPTIONAL: This defaults to the namespace where this CR exists
+      encryptionKeySecretNamespace: unifiedpush
```
camilamacedo86: the namespace is not required and can b obtained programmatically by instance.Namespace

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -30,6 +30,51 @@ spec:
...
+                  name:
+                    description: Name is the name that will be given to the resulting
+                      CronJob
+                    type: string
+                  schedule:
+                    description: Schedule is the schedule that the job will be run
+                      at, in cron format
+                    type: string
+                required:
+                - name
```
camilamacedo86, line 67: I liked the descriptions. 👍 
I just believe that the name, labels, and schedule is not required.

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
```diff
@@ -4,15 +4,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // UnifiedPushServerSpec defines the desired state of UnifiedPushServer
 // +k8s:openapi-gen=true
 type UnifiedPushServerSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
camilamacedo86, line 13: It is missing the CR Status. How the admin/user can check the data used in the case of failures?
Following an example over the impl here : https://github.com/aerogear/mobile-security-service-operator/pull/125 (Note that  added XXX in the fields
Screenshot 2019-06-28 at 09 27 28
 just to not expose them here)

pkg/config/config.go
camilamacedo86: @grdryn in all my tests it worked without the need of this logic. If the ENV VAR of the encryption secret is == "" it worked and is not required the if at all.

pkg/controller/unifiedpushserver/backups.go
secondsun: Are secret names limited to 63 characters?

odra: all resources names (and labels) are limited to 63 chars: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

camilamacedo86: @grdryn if someone changes the ENV VARs values manually by OCP console it will stop to work. To avoid this scenario in the MSS Oper we are checking the ENV VAR values from the database pod as the service name dynamically and creating all programmatically. 

c/c @odra

grdryn: This is a good point. I think if we can work out (based on the suffixes that we append) what max string lengths for names that we can have that we should add to our custom kinds. I think adding kubebuilder validation annotations should cause ones that are over that to be rejected, but I'm not sure if that's enabled in OpenShift yet or not.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
grdryn: This supports a list of backups, rather than just one, which means that if we use `instance.Name`, we'd also need to have a generated part (or accept a suffix here). I think it's nicer to allow humans to give a name that can have some relation to what kind of backup it is (either based on what it's backing up (database, resources, etc) or the schedule, or whatever). Also, this is mostly copying the keycloak-operator.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
grdryn: Yes, labels are not required. These are just optional labels that someone can add (yeah, this 'monitoring-key' example one is probably a bad example, again just copied from the keycloak operator).

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
grdryn: I would be fine with _just_ passing those values in the CR, or _just_ passing the secret name like here. I'm not keen to have both ways though, and I'm mostly doing it this way because of how the keycloak operator does it.

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
grdryn: They could just look in the secrets that they've specified in the spec section.

pkg/config/config.go
grdryn: Going to change this to 10 for now as discussed in hangouts.

@camilamacedo86 your most recent comment here doesn't look like it's related to this line. Is it referring to some other part of the change?

pkg/config/config.go
grdryn: We had an earlier discussion here somewhere about that, and decided to pin it to a specific version, to avoid surprises.

pkg/controller/unifiedpushserver/backups.go
grdryn: I think it's mostly because it's already populated with some values (line 16), and this is just adding any optional other ones that have been specified in the CR for this particular backup

pkg/controller/unifiedpushserver/backups.go
grdryn: Again, this is following how it's done in the keycloak operator. I like the idea of using a different ServiceAccount here, since it's for a different purpose (that has different access requirements). For example, this ServiceAccount needs to be able to read Secrets in whatever namespaces the Secrets it needs are placed in.

pkg/controller/unifiedpushserver/backups.go
grdryn: Good catch! I'll change that. :+1:

pkg/controller/unifiedpushserver/backups.go
camilamacedo86: @grdryn the database has env vars (user/pwd/database name) with the values usually. I think it is the way that was impl here as well. So it means that a user can change it for someone reason. ( by OCP UI for example ) 

In MSS oper we fetch the pod database then get the env vars used. You can check here we building the DB secret data . https://github.com/aerogear/mobile-security-service-operator/pull/125/files#diff-0a376a010c2ec86f696671db622a8955R12

grdryn: @camilamacedo86 yes, for the container they're passed as env vars, but those env var values are taken from the secret, so I don't think we suffer from the issue that you're describing. Also, since it's just one controller for all of this, it should be kept in sync, even if someone does change something.

Here's where we create the secret:
https://github.com/aerogear/unifiedpush-operator/blob/076d019ad3c21a3ed82ad505e158e51f71aa275b/pkg/controller/unifiedpushserver/postgresql.go#L41..L50

Here's where we link the secret values to the container env vars: https://github.com/aerogear/unifiedpush-operator/blob/076d019ad3c21a3ed82ad505e158e51f71aa275b/pkg/controller/unifiedpushserver/postgresql.go#L88..L122

pkg/config/config.go
camilamacedo86: It is true. Sorry. It was over the if to do the args which are passed to the container in the cronJob.

pkg/controller/unifiedpushserver/backups.go
camilamacedo86: Hi @grdryn,

After the installation via the UPS oper, all will be fine. However, if the admin changed it in OCP for any reason the backup will stop to work. IMHO: We can avoid many issue scenarios by getting the data dynamically. It will be used now for the RHMI managed by us and self-managed as well. Also, by the community which should be able to use just the UPS oper alone to get the service as too. But we can improve it latter too.

pkg/config/config.go
camilamacedo86: @grdryn it is fine. I spoke with Leo and changed it for the fixed version too. 
I just think that it should be done in the CR. See that:

RHMI can setup it now with 1.0.8
But tomorrow we can have a MobileX which will require the version 2.0.0
Each platform will have one release date and process.

So, the appropriate place for any configuration regards the CRD in POV is in its CR. The operator Env Var
should have things that are relevant to the operator itself. For example, the namespaces which can be used for it, its image, its level of logs, etc..

pkg/controller/unifiedpushserver/backups.go
camilamacedo86: I think it is not required at all but I am closing it since we can improve it latter too.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
camilamacedo86: I think is not required at all. We can remove this part and make it more simple if you wish.

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
camilamacedo86: @grdryn if something goes wrong and it is not using what was spec then you will not able to check it. It is the point. Also, see that many of data should/could be generated dynamically and programmatically.  IMHO: The cronJobStatus it's important to.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
camilamacedo86: I am closing it. yes is required the namespace you are correct. we need to do it in the MSS Oper too

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
camilamacedo86: I am closing since we can improve it latter.

pkg/controller/unifiedpushserver/backups.go
camilamacedo86: OKs.

pkg/config/config.go
camilamacedo86: Anyway, we can improve it after so I am closing it. I just would like to share the idea with you.

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
camilamacedo86: We can do it after too if you wish.

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
grdryn: Yes, I think you're right, I'll remove this labels section now.

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
grdryn: I've removed the labels, but I believe the name and schedule should always be specified.

grdryn:
I've added a couple of new changes here. The most notable is proper reconciliation of the CronJob resources (add/delete/update, based on changes to the UPS CR).

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_backup.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: push.aerogear.org/v1alpha1
```
camilamacedo86, line 1: Hi @grdryn why you duplicated the  UnifiedPushServer instead of just create a new CR/CRD for the bkp?

Note that in the way that you did here the user cannot decide to apply the bkp after without re-creating all.

grdryn: I didn't duplicate the UnifiedPushServer, I just provided an additional example CR. Now there's an example with backups in addition to the pre-existing example without backups.

A user _can_ add a backup (or remove or update one) without re-creating anything, since the backups get reconciled properly now (see the most recent commit).

camilamacedo86: If I have the UPS installed then I would like to have the bkp, should I not remove the CR(service) then re-apply it with the bkp details as the new example? If yes, it does not means that the service will be deleted and re-created?

IHMO it should/could be encapsulated as it is done for the other operators in the operatorhub.io. See for example etcd https://operatorhub.io/operator/etcd

pkg/apis/push/v1alpha1/iosvariant_types.go
```diff
@@ -45,6 +45,10 @@ type IOSVariantStatus struct {
 
 // IOSVariant is the Schema for the iosvariants API
 // +k8s:openapi-gen=true
+// +kubebuilder:printcolumn:name="Production",type="boolean",JSONPath=".spec.production"
```
aliok, line 48: @grdryn generating these stuff is great!

aliok (approved):
I did the cronjob creation verification. 
The actual backup and upload to AWS verification --> as you said Leo did it and that's enough.

All good. Feel free to merge.

PR #40, opened 2019-07-09
AEROGEAR-9024  Expose variant secret in CR status

jomrazek:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9024

## What
Expose secret field for both variants

## Verification Steps
Create android/ios variant check secret for correct data

## Checklist:

- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/apis/push/v1alpha1/androidvariant_types.go
+++ b/pkg/apis/push/v1alpha1/androidvariant_types.go
@@ -37,6 +37,7 @@ type AndroidVariantSpec struct {
 type AndroidVariantStatus struct {
 	Ready     bool   `json:"ready"`
 	VariantId string `json:"variantId,omitempty"`
+	Secret    string `json:"secret,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/controller/androidvariant/androidvariant_controller.go
+++ b/pkg/controller/androidvariant/androidvariant_controller.go
@@ -122,13 +122,14 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, nil
 	}
 
-	variantId, err := unifiedpushClient.CreateAndroidVariant(instance)
+	variantId, secret, err := unifiedpushClient.CreateAndroidVariant(instance)
 	if err != nil {
 		reqLogger.Error(err, "Error creating Android variant in UPS.", "AndroidVaraint.Name", instance.Name)
 		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
 	instance.Status.VariantId = variantId
+	instance.Status.Secret = secret
 	instance.Status.Ready = true
 	err = r.client.Status().Update(context.TODO(), instance)
 	if err != nil {
--- a/pkg/unifiedpush/client.go
+++ b/pkg/unifiedpush/client.go
@@ -159,7 +159,7 @@ func (c UnifiedpushClient) GetAndroidVariant(v *pushv1alpha1.AndroidVariant) (st
 }
 
 // CreateAndroidVariant creates a Variant on an Application in UPS
-func (c UnifiedpushClient) CreateAndroidVariant(v *pushv1alpha1.AndroidVariant) (string, error) {
+func (c UnifiedpushClient) CreateAndroidVariant(v *pushv1alpha1.AndroidVariant) (string, string, error) {
 	url := fmt.Sprintf("%s/rest/applications/%s/android", c.Url, v.Spec.PushApplicationId)
 
 	params := map[string]string{
@@ -171,7 +171,7 @@ func (c UnifiedpushClient) CreateAndroidVariant(v *pushv1alpha1.AndroidVariant) 
 
 	payload, err := json.Marshal(params)
 	if err != nil {
-		return "", errors.Wrap(err, "Failed to marshal android variant params to json")
+		return "", "", errors.Wrap(err, "Failed to marshal android variant params to json")
 	}
 
 	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(payload))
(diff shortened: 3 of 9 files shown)


pkg/controller/androidvariant/configmap.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	pushv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *ReconcileAndroidVariant) createAndroidConfigMap(androidVariant *pushv1alpha1.AndroidVariant) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "android-variant-configmap",
```
grdryn: Change the value here to `androidVariant.Name` -- we need to be able to support more than one variant.

pkg/controller/androidvariant/configmap.go
```diff
@@ -0,0 +1,26 @@
...
+			Namespace: "mobile-developer-console",
```
grdryn: Change this to `androidVariant.Namespace` -- the fact that for our use-case it will be mobile-developer-console is not that relevant, we should be creating it wherever the CR was created.

pkg/controller/androidvariant/androidvariant_controller.go
```diff
@@ -140,6 +148,32 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
...
+		}
+
+		// ConfigMap created successfully - don't requeue
+		return reconcile.Result{}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// ConfigMap already exists - don't requeue
+	reqLogger.Info("Skip reconcile: ConfigMap already exists", "ConfigMap.Namespace", found.Namespace, "ConfigMap.Name", found.Name)
```
grdryn: I think this message and the comment above it should be removed. They're a bit misleading.

pkg/controller/androidvariant/configmap.go
```diff
@@ -0,0 +1,26 @@
...
+		},
+		Data: createConfigMapData(androidVariant),
+	}
+}
+
+func createConfigMapData(androidVariant *pushv1alpha1.AndroidVariant) map[string]string {
+	config := map[string]string{
```
grdryn: @ziccardi Is the information in this ConfigMap sufficient for your needs?

pkg/controller/androidvariant/configmap.go
```diff
@@ -0,0 +1,26 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (r *ReconcileAndroidVariant) createAndroidConfigMap(androidVariant *pushv1alpha1.AndroidVariant) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "android-variant-configmap",
+			Namespace: "mobile-developer-console",
+		},
+		Data: createConfigMapData(androidVariant),
```
grdryn:
```suggestion
		Data: map[string]string{
			"senderId":      androidVariant.Spec.SenderId,
			"variantId":     androidVariant.Status.VariantId,
			"variantSecret": androidVariant.Spec.ServerKey,
		}
```

^this removes the need for the `createConfigMapData` function.

grdryn (changes requested):
@jomrazek the error you were getting when trying to test this out is because you're running the operator locally, where it can't reach the `example-unifiedpushserver-unifiedpush` service in the cluster. You need to port-forward that out to your local machine. There's an example of how to do that in the readme.adoc. If you can't get that to work for some reason related to your machine, then just build the operator image and deploy it into the cluster as it would be run in production.

pkg/unifiedpush/client.go
```diff
@@ -253,10 +253,10 @@ func (c UnifiedpushClient) CreateIOSVariant(v *pushv1alpha1.IOSVariant) (string,
 		"production":  strconv.FormatBool(v.Spec.Production),
 	}
 
-	// We need to decode it before sending
+	// We need tpkg/controller/iosvariant/configmap.goo decode it before sending
```
grdryn: This looks like a stray ctrl+v or something :smiley_cat:

pkg/apis/push/v1alpha1/androidvariant_types.go
```diff
@@ -35,8 +35,9 @@ type AndroidVariantSpec struct {
 // AndroidVariantStatus defines the observed state of AndroidVariant
 // +k8s:openapi-gen=true
 type AndroidVariantStatus struct {
-	Ready     bool   `json:"ready"`
-	VariantId string `json:"variantId,omitempty"`
+	Ready         bool   `json:"ready"`
+	VariantId     string `json:"variantId,omitempty"`
+	VariantSecret string `json:"variantSecret,omitempty"`
```
grdryn: Given that this is called `Secret` in UPS, rather than `VariantSecret`, I think we should have it as `Secret` here too.

grdryn (commented):
We may not need the configmap stuff now (pending confirmation), but we'll definitely need the secret field in the status.

pkg/apis/push/v1alpha1/androidvariant_types.go
jomrazek: I don't mind changing it, but it's variant secret in mdc and also in ups ui

aliok:
> We may not need the configmap stuff now (pending confirmation), but we'll definitely need the secret field in the status.

@grdryn @wei-lee 
Can we now say that we won't do the configmap stuff for sure?

grdryn:
@aliok Yes. We spoke about it yesterday and all agreed. I created new issues for the other places where we're creating configmaps from other operators to get them removed (and MDC updated to use the status).

grdryn:
@jomrazek the title of this PR has [WIP] in it. Is that still accurate?

jomrazek:
@grdryn nope, fixed

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -8,8 +8,6 @@ spec:
     kind: UnifiedPushServer
     listKind: UnifiedPushServerList
     plural: unifiedpushservers
-    shortNames:
```
grdryn: This should not be removed. What version of the `operator-sdk` do you have? You may need to update to a non-released version.

grdryn (commented):
I've verified this. Once it's rebased and the shortname is added back, I'll approve and it can be merged :+1:

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
jomrazek: 0.7.0

grdryn: Could you try the latest (0.8.1 I think), and if this is still getting removed, installing from master will fix that.

grdryn (approved):

PR #42, opened 2019-07-10
verify the provided namespace before continuing to reconcile

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9568

## Image
`quay.io/aerogear/unifiedpush-operator:f144d9e`

## What
Only respond to consumption crs which are created in a namespace which exists in a user specified list or in the oeprator namespace

## How
Check that the namespace is in the ENV VAR - APP_NAMESPACES or it is the operator namespace before moving on with the reconcile

## Verification Steps
`make cluster/clean`
`make install`
Wait for ups to come up

**Verify operator namespace**
_Steps_
kubectl apply -n unifiedpush -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 
_Expected result_
Application is created in ups console
_Cleanup_
kubectl delete -n unifiedpush -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 

**Verify APP_NAMESPACES namespace**
_Steps_
kubectl apply -n unifiedpush-apps -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 
_Expected result_
Application is created in ups console
_Cleanup_
kubectl delete -n unifiedpush-apps -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 

**Verify neither of the above namespace**
_Steps_
oc new-project test-ups
kubectl apply -n test-ups -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 
_Expected result_
Application is **not created** in ups console
_Cleanup_
kubectl delete -n test-ups -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml 
oc delete project ups-test

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ APP_NAMESPACES=unifiedpush-apps
 CODE_COMPILE_OUTPUT = build/_output/bin/unifiedpush-operator
 TEST_COMPILE_OUTPUT = build/_output/bin/unifiedpush-operator-test
 
+DEV_TAG ?= $(shell sh -c "git rev-parse --short HEAD")
+
 ##############################
 # Local Development          #
 ##############################
@@ -163,20 +165,10 @@ cluster/clean:
 	- kubectl delete namespace ${APP_NAMESPACES}
 
 
-.PHONY: image/build/master
-image/build/master:
-	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:master
-
- .PHONY: image/push/master
-image/push/master:
-	docker push quay.io/${ORG_NAME}/${APP_NAME}:master
-
-
-
- .PHONY: image/build/dev
-image/build/dev:
+.PHONY: image/build
+image/build:
 	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:${DEV_TAG}
 
- .PHONY: image/push/dev
-image/push/dev:
+.PHONY: image/push
+image/push:
 	docker push quay.io/${ORG_NAME}/${APP_NAME}:${DEV_TAG}
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TEST_PKGS = $(addprefix $(PKG)/,$(PACKAGES))
 APP_FILE=./cmd/manager/main.go
 
 NAMESPACE=unifiedpush
+APP_NAMESPACES=unifiedpush-apps
 CODE_COMPILE_OUTPUT = build/_output/bin/unifiedpush-operator
 TEST_COMPILE_OUTPUT = build/_output/bin/unifiedpush-operator-test
 
@@ -101,24 +102,48 @@ monitoring/uninstall:
 .PHONY: example-pushapplication/apply
 example-pushapplication/apply:
 	@echo ....... Applying the PushApplication example in the current namespace  ......
-	- kubectl apply -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+	- kubectl apply -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
 
 .PHONY: example-pushapplication/delete
 example-pushapplication/delete:
 	@echo ....... Deleting the PushApplication example in the current namespace  ......
-	- kubectl delete -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+	- kubectl delete -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+
+.PHONY: android-variant/apply
+android-variant/apply:
+	make example-pushapplication/apply
+	@echo ....... Applying the Android Variant example in the current namespace  ......
+	- kubectl apply -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_androidvariant_cr.yaml
+
+.PHONY: android-variant/delete
+android-variant/delete:
+	@echo ....... Applying the Android Variant example in the current namespace  ......
+	- kubectl delete -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_androidvariant_cr.yaml
+
+
+.PHONY: ios-variant/apply
+ios-variant/apply:
+	make example-pushapplication/apply
+	@echo ....... Applying the Android Variant example in the current namespace  ......
+	- kubectl apply -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_iosvariant_cr.yaml
+
+.PHONY: ios-variant/delete
+ios-variant/delete:
+	@echo ....... Applying the Android Variant example in the current namespace  ......
+	- kubectl delete -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_iosvariant_cr.yaml
 
 .PHONY: cluster/prepare
 cluster/prepare:
 	- kubectl create namespace $(NAMESPACE)
 	- kubectl label namespace $(NAMESPACE) monitoring-key=middleware
+	- kubectl create namespace $(APP_NAMESPACES)
 	- kubectl create -n $(NAMESPACE) -f deploy/service_account.yaml
 	- kubectl create -n $(NAMESPACE) -f deploy/role.yaml
 	- kubectl create -n $(NAMESPACE) -f deploy/role_binding.yaml
-	- kubectl apply -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_pushapplication_crd.yaml
-	- kubectl apply -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_androidvariant_crd.yaml
-	- kubectl apply -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_iosvariant_crd.yaml
-	- kubectl apply -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
+	- kubectl apply -f deploy/crds/push_v1alpha1_pushapplication_crd.yaml
+	- kubectl apply -f deploy/crds/push_v1alpha1_androidvariant_crd.yaml
+	- kubectl apply -f deploy/crds/push_v1alpha1_iosvariant_crd.yaml
+	- kubectl apply -f deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
 
 .PHONY: cluster/clean
 cluster/clean:
@@ -129,9 +154,29 @@ cluster/clean:
 	- kubectl delete -f deploy/role.yaml
 	- kubectl delete -n $(NAMESPACE) -f deploy/role_binding.yaml
 	- kubectl delete -n $(NAMESPACE) -f deploy/service_account.yaml
-	- kubectl delete -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_pushapplication_crd.yaml
-	- kubectl delete -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_androidvariant_crd.yaml
-	- kubectl delete -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_iosvariant_crd.yaml
-	- kubectl delete -n $(NAMESPACE) -f deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
+	- kubectl delete -f deploy/crds/push_v1alpha1_pushapplication_crd.yaml
+	- kubectl delete -f deploy/crds/push_v1alpha1_androidvariant_crd.yaml
+	- kubectl delete -f deploy/crds/push_v1alpha1_iosvariant_crd.yaml
+	- kubectl delete -f deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
 	- make monitoring/uninstall
 	- kubectl delete namespace $(NAMESPACE)
+	- kubectl delete namespace ${APP_NAMESPACES}
+
+
+.PHONY: image/build/master
+image/build/master:
+	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:master
+
+ .PHONY: image/push/master
+image/push/master:
+	docker push quay.io/${ORG_NAME}/${APP_NAME}:master
+
+
+
+ .PHONY: image/build/dev
+image/build/dev:
+	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:${DEV_TAG}
+
+ .PHONY: image/push/dev
+image/push/dev:
+	docker push quay.io/${ORG_NAME}/${APP_NAME}:${DEV_TAG}
--- a/pkg/controller/androidvariant/androidvariant_controller.go
+++ b/pkg/controller/androidvariant/androidvariant_controller.go
@@ -89,9 +89,10 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-	err = util.IsValidAppNamespace(instance.Namespace, operatorNamespace, instance.Name)
-	if err != nil {
-		return reconcile.Result{}, err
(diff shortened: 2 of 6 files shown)


aliok:
@laurafitzgerald 
A push app was created in UPS console in the 3rd command as well:
```

{"level":"info","ts":1562752746.5297067,"logger":"controller_pushapplication","msg":"Reconciling PushApplication","Request.Namespace":"test-ups","Request.Name":"example-pushapplication"}
--
  | {"level":"info","ts":1562752746.5491757,"logger":"controller_pushapplication","msg":"Reconciling PushApplication","Request.Namespace":"test-ups","Request.Name":"example-pushapplication"}
  | {"level":"info","ts":1562752773.4841197,"logger":"controller_pushapplication","msg":"Reconciling PushApplication","Request.Namespace":"test-ups","Request.Name":"example-pushapplication"}
  | {"level":"info","ts":1562752773.5069547,"logger":"controller_pushapplication","msg":"Adding Finalizer to the PushApplication","Request.Namespace":"test-ups","Request.Name":"example-pushapplication"}
  | {"level":"info","ts":1562752773.5103362,"logger":"controller_pushapplication","msg":"Push app created","Request.Namespace":"test-ups","Request.Name":"example-pushapplication","PushApp.Name":"example-pushapplication","PushApp.Id":"f55e952d-c0db-48d8-9235-b6eb0b1faf9a","PushApp.MasterSecret":"bf14bc4c-d701-4007-bb14-ac336161c122"}
```

aliok:
@laurafitzgerald works as expected.

However, this error message is written every few seconds forever:
```
{"level":"error","ts":1562754263.776352,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"pushapplication-controller","request":"test-ups/example-pushapplication","error":"The app cr example-pushapplication was created in a namespace which is not present in the APP_NAMESPACES environment variable provided to the operator","stacktrace":"github.com/aerogear/unifiedpush-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/aerogear/unifiedpush-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:217\ngithub.com/aerogear/unifiedpush-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158\ngithub.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/home/aliok/go/src/github.com/aerogear/unifiedpush-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
```

I think the ideal way to solve this problem is writing the error message only once and creating a Kube event with this problem.
I don't have any experience with this though. For now perhaps just print it once and mark the CR status as failed and don't print this error message again and again?

Let's see what operators master says: @odra

Makefile
```diff
@@ -129,9 +154,29 @@ cluster/clean:
...
\ No newline at end of file
+	- kubectl delete namespace $(NAMESPACE)
+	- kubectl delete namespace ${APP_NAMESPACES}
+
+
+.PHONY: image/build/master
+image/build/master:
+	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:master
+
+ .PHONY: image/push/master
```
grdryn: I'd remove these tasks to build & push master. The Jenkins pipeline does that already, and we don't want people to think they can overwrite it with an image they built on their own machine.

Makefile
```diff
@@ -129,9 +154,29 @@ cluster/clean:
...
+image/push/master:
+	docker push quay.io/${ORG_NAME}/${APP_NAME}:master
+
+
+
+ .PHONY: image/build/dev
+image/build/dev:
+	operator-sdk build quay.io/${ORG_NAME}/${APP_NAME}:${DEV_TAG}
```
grdryn, line 170: Is `${DEV_TAG}` set at this point? I can't see it. Maybe something like `DEV_TAG ?= $(shell sh -c "git rev-parse --short HEAD")` at the top of the file is needed?

Also given that there will not be an `image/build/master` and `image/push/master`, maybe the `/dev` can be removed from these task names. Wdyt?

pkg/controller/pushapplication/pushapplication_controller.go
```diff
@@ -81,6 +82,17 @@ func (r *ReconcilePushApplication) Reconcile(request reconcile.Request) (reconci
 		return reconcile.Result{}, err
 	}
 
+	// Check if the namespace of the cr is in the APP_NAMESPACES environment variable provided to the operator or if it's in the operator namespace
+	operatorNamespace, err := k8sutil.GetOperatorNamespace()
```
grdryn, line 86: Does that `k8sutil.GetOperatorNamespace()` return something valid when it's run locally? If not, does it return an error?

pkg/controller/util/helpers.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+import (
+	"os"
+	"strings"
+)
+
+// IsValidAppNamespace returns an error when the namespace passed is not present in the APP_NAMESPACES environment variable provided to the operator.
+func IsValidAppNamespace(crnamespace string, operatorNamespace string, name string) bool {
```
grdryn:
```suggestion
func IsValidAppNamespace(crNamespace string, operatorNamespace string, name string) bool {
```

Makefile
laurafitzgerald: 👍

laurafitzgerald:
Thanks @grdryn Changes pushed to address your feedback.

pkg/controller/pushapplication/pushapplication_controller.go
aliok: Didn't really try it but I don't think it would work when running locally:

```
// GetOperatorNamespace returns the namespace the operator should be running in.
func GetOperatorNamespace() (string, error) {
	nsBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
	if err != nil {
		if os.IsNotExist(err) {
			return "", ErrNoNamespace
		}
		return "", err
	}
	ns := strings.TrimSpace(string(nsBytes))
	log.V(1).Info("Found namespace", "Namespace", ns)
	return ns, nil
}
```

aliok: I think we can use `k8sutil.GetWatchNamespace()` here as it is already used in `main.go`.

That can be used for sure.

https://github.com/aerogear/unifiedpush-operator/blob/f144d9ed14b0db7987a51d62dc2a31f2dc745547/cmd/manager/main.go#L67

grdryn: If we run the operator locally, that `k8sutil.GetOperatorNamespace()` will return with that `ErrNoNamespace` error, because the file it checks doesn't exist. This means that if we try to create a PushApplication in the unifiedpush namespace, it won't work.

We currently set the watch namespace to be an empty string so that the operator will watch more than one namespace (it needs to watch its own namespace, as that's where we want the UPS instance to be, and it also needs to watch the namespace(s) where PushApplication and Variant CRs will be created). When running `operator-sdk up local` we could add `--namespace unifiedpush`, and that will set the watch namespace to that value, but again, that will cause the operator to only listen in that one namespace I think. Maybe that's sufficient when running the operator locally?

Alternatively, if `k8sutil.GetOperatorNamespace()` returns an error, we could have it check some undocumented env var such as `_LOCAL_OPERATOR_NAMESPACE` or something, and pass that in the Makefile?

aliok: I am confused. There are too many namespace contexts here :) Watch, operator, app, anything else? :)

Let me think out loud.

Let's go 1 step back.

* We want the UPS operator to run in unifiedpush namespace
* We want the UPS operator to create UPS server in unifiedpush namespace
* We want the UPS operator to expect the CRs in `APP_NAMESPACES`
** Does it also include the installation CR? Or is it only for service consumption CRs?
* We want he UPS operator to watch for pods/deployments/etc. in unifiedpush namespace.

We already have a separate env var for `APP_NAMESPACES`.
For all the others, we just need to use unifiedpush.
* Running in cluster: `k8sutil.GetOperatorNamespace()` should work.
* Running locally: Using `WATCH_NAMESPACE` can be enforced in the makefile.

So, here:

https://github.com/aerogear/unifiedpush-operator/pull/42#discussion-diff-302098737R86

 we can do like:
* Get `k8sutil.GetOperatorNamespace()`.
* If it fails, get `k8sutil.GetWatchNamespace()`
* If it is empty, `panic`

wdyt?

grdryn: My understanding of the problem is that:

* We have `APP_NAMESPACES` because we want to limit the namespaces that we react to consumption CRs in, but we also want to react to them in the operator namespace, whether it's part of `APP_NAMESPACES` or not (a user should be able to do everything with this operator in one namespace if they want). To answer your other question about `APP_NAMESPACES`, it's only in the "consumption" controllers that we want to check this in (the whole idea of it is a dirty hack, since the operator will react to "consumption" CRs in any namespace, but will just refuse to do anything if it's not in an acceptable namespace).

* We need to set `WATCH_NAMESPACE` to an empty string to generally allow listening in other namespaces than the operator one.

So the problem with your proposed steps is that if the `WATCH_NAMESPACE` env var is not an empty string, and you have anything other than the same single namespace in `APP_NAMESPACES`, those additional namespaces will be ignored.

Maybe it's fine for now that for _local_ development, that we disallow creation of "consumption" CRs (not sure how I feel about that term), in the operator namespace, until there's a better solution from the operator-sdk (I think there will be)? What do you think? :thinking:

aliok: > "consumption" CRs (not sure how I feel about that term)

LOL :) I didn't make it up, somebody else did :grin: 

> What do you think? 

Sounds good. Let's proceed with this.

grdryn: >> "consumption" CRs (not sure how I feel about that term)
>
> LOL :) I didn't make it up, somebody else did :grin: 

I know, and I use it sometimes too; it was just in my third usage of it in that comment I realised I'd had enough of it :sweat_smile: 

>> What do you think?
>
> Sounds good. Let's proceed with this.

Cool, I think that means that this PR is already as expected. I'll just double-check.

Makefile
```diff
@@ -101,24 +104,48 @@ monitoring/uninstall:
...
 example-pushapplication/delete:
 	@echo ....... Deleting the PushApplication example in the current namespace  ......
-	- kubectl delete -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+	- kubectl delete -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+
+.PHONY: android-variant/apply
+android-variant/apply:
+	make example-pushapplication/apply
+	@echo ....... Applying the Android Variant example in the current namespace  ......
+	- kubectl apply -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_androidvariant_cr.yaml
```
grdryn, line 118: I think if we're going to have this make target here, it should actually apply an android variant with a populated pushApplicationId.

OTOH, While the above should be possible, we might be going a little make-crazy will all of these new things.

Makefile
```diff
@@ -101,24 +104,48 @@ monitoring/uninstall:
...
 
 .PHONY: example-pushapplication/delete
 example-pushapplication/delete:
 	@echo ....... Deleting the PushApplication example in the current namespace  ......
-	- kubectl delete -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+	- kubectl delete -n $(APP_NAMESPACES) -f deploy/crds/examples/push_v1alpha1_pushapplication_cr.yaml
+
+.PHONY: android-variant/apply
+android-variant/apply:
+	make example-pushapplication/apply
```
grdryn, line 116: This can be set as a pre-requisite to the target, rather than invoking make again. E.g. on the line above:

```
android-variant/apply: example-pushapplication/apply
```

Same with the other targets that have dependencies.

pkg/controller/androidvariant/androidvariant_controller.go
```diff
@@ -83,6 +84,17 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// Check if the namespace of the cr is in the APP_NAMESPACES environment variable provided to the operator or if it's in the operator namespace
+	operatorNamespace, err := k8sutil.GetOperatorNamespace()
```
grdryn, line 88: This, and the resulting error check immediately below, _could_ be part of the `util.IsValidAppNamespace()` function, since in all 3 controllers we're calling this just to pass the result to that function.

pkg/controller/androidvariant/androidvariant_controller.go
```diff
@@ -83,6 +84,17 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	is := util.IsValidAppNamespace(instance.Namespace, operatorNamespace, instance.Name)
+	if !is {
+		reqLogger.Info("The app cr %s was created in a namespace which is not present in the APP_NAMESPACES environment variable provided to the operator or is not the in the operator namespace", instance.Name)
```
grdryn, line 94:
```suggestion
		reqLogger.Info(fmt.Sprintf("The app cr %s was created in a namespace which is not present in the APP_NAMESPACES environment variable provided to the operator or is not the in the operator namespace", instance.Name))
```

This should fix the error that `reqLogger.Info` doesn't have the correct number of parameters, and also populates the `%s` in the format string as expected.

pkg/controller/util/helpers.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// IsValidAppNamespace returns an error when the namespace passed is not present in the APP_NAMESPACES environment variable provided to the operator.
+func IsValidAppNamespace(crNamespace string, operatorNamespace string, name string) bool {
+	appNamespacesEnvVar, found := os.LookupEnv("APP_NAMESPACES")
+	if !found {
+		return false
+	}
+
+	if operatorNamespace == crNamespace {
```
grdryn, line 15: I would put this above the check for "APP_NAMESPACES", since if that's not set, CRs in the operator namespace should still be considered valid.

grdryn (changes requested):
Some small comments inline. Some of those should be done here I think, and others are more like suggestions. Generally a good approach though! :+1:

camilamacedo86 (changes requested):
@laurafitzgerald I'd recommend you do as it was applied here: https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-fb81b0a96eed5b010cfb081b3aef4431R12 and here https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-2752e78fa6777e8cd072967fb0801b5fR99. Following the reasons, 

1. Allow we still running the operator locally ( outside of the cluster ) 
2. Allow we impl unit test.

c/c @grdryn @aliok

grdryn:
Closing this in favour of #46.

PR #46, opened 2019-07-15
Only watch relevant namespaces for CRs

grdryn:
Before this change, we'd try to reconcile a CR in any namespace. This
changes that so that the controller doesn't even bother calling the
Reconcile method if the CR isn't in one of the namespaces that we
expect for that kind.

Since running the operator locally with operator-sdk up local doesn't
seem to correctly react to CRs in namespaces other than one for the
current context, this change also adds that to the list of
APP_NAMESPACES for the `make code/run` target.

### DIFF
--- a/pkg/controller/iosvariant/iosvariant_controller.go
+++ b/pkg/controller/iosvariant/iosvariant_controller.go
@@ -6,6 +6,7 @@ import (
 
 	pushv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
 	"github.com/aerogear/unifiedpush-operator/pkg/controller/util"
+	"github.com/aerogear/unifiedpush-operator/pkg/nspredicate"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -42,7 +43,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to primary resource IOSVariant
-	err = c.Watch(&source.Kind{Type: &pushv1alpha1.IOSVariant{}}, &handler.EnqueueRequestForObject{})
+	onlyEnqueueForValidNamespaces, err := nspredicate.NewFromEnvVar("APP_NAMESPACES")
+	if err != nil {
+		return err
+	}
+	err = c.Watch(
+		&source.Kind{Type: &pushv1alpha1.IOSVariant{}},
+		&handler.EnqueueRequestForObject{},
+		onlyEnqueueForValidNamespaces,
+	)
 	if err != nil {
 		return err
 	}
--- a/pkg/nspredicate/nspredicate.go
+++ b/pkg/nspredicate/nspredicate.go
@@ -1,0 +1,48 @@
+package nspredicate
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// NewNamespacePredicateFromEnvVar looks up the provided environment
+// variable, parses the value, and returns a NamespacePredicate
+func NewFromEnvVar(envVar string) (NamespacePredicate, error) {
+	namespaces, found := os.LookupEnv(envVar)
+	if !found {
+		return NamespacePredicate{}, fmt.Errorf("Environment variable %s not found", envVar)
+	}
+	return NamespacePredicate{Namespaces: strings.Split(namespaces, ",")}, nil
+}
+
+type NamespacePredicate struct {
+	Namespaces []string
+}
+
+func (p NamespacePredicate) Create(e event.CreateEvent) bool {
+	return p.isValidNamespace(e.Meta.GetNamespace())
+}
+
+func (p NamespacePredicate) Delete(e event.DeleteEvent) bool {
+	return p.isValidNamespace(e.Meta.GetNamespace())
+}
+
+func (p NamespacePredicate) Update(e event.UpdateEvent) bool {
+	return p.isValidNamespace(e.MetaOld.GetNamespace())
+}
+
+func (p NamespacePredicate) Generic(e event.GenericEvent) bool {
+	return p.isValidNamespace(e.Meta.GetNamespace())
+}
+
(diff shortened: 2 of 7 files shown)


pkg/controller/iosvariant/iosvariant_controller.go
```diff
@@ -42,7 +43,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to primary resource IOSVariant
-	err = c.Watch(&source.Kind{Type: &pushv1alpha1.IOSVariant{}}, &handler.EnqueueRequestForObject{})
+	onlyEnqueueForValidNamespaces, err := nspredicate.NewFromEnvVar("APP_NAMESPACES")
```
camilamacedo86, line 46: The implementation needs to be fone in the reconcile. The WATCH is just observers(triggers) in the cluster. It will not perform what is required here. Note that the solution applied in the MSS Oper is the same applied for other old projects maintained by us. See [here](https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-2752e78fa6777e8cd072967fb0801b5fR99) how it was done in the PR for App Metrics.  

When we run `operator-sdk local up` or `unit-test` it is outside of the cluster and because of this, a specific error is faced when we call the framework features. Because of this, we are able to identify these scenarios and allow continue. See [here](https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-3661dc32b6bde1b0939ad409364e48fcR1) its tests to ensure that works well as its implementation [here](https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-fb81b0a96eed5b010cfb081b3aef4431R65)

grdryn: This seems like a much better place to check than in the reconcile method, and exactly the kind of thing that this is for. It works exactly as I expect (both inside and outside of the cluster, with the caveat that locally it seems to only be able to react to events in one namespace, but that was the case beforehand also).

> It will not perform what is required here.

Please elaborate. What do you see as being "required" here that will not be "performed"?

> it is outside of the cluster and because of this, a specific error is faced when we call the framework features.

Please elaborate. What "framework features" are being used here that won't work when running locally?

camilamacedo86: 👍 You added the validation in the primary one. Yep. It should work in the **cluster**. You are correct.

**However, note that it is not working for the local tests.** 

```shell
{"level":"error","ts":1563273007.010122,"logger":"cmd","msg":"","error":"Environment variable APP_NAMESPACES not found","stacktrace":"github.com/aerogear/unifiedpush-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/camilamacedo/go/src/github.com/aerogear/unifiedpush-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/Users/camilamacedo/go/src/github.com/aerogear/unifiedpush-operator/cmd/manager/main.go:127\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:201"}
Error: failed to run operator locally: (failed to exec []string{"build/_output/bin/unifiedpush-operator-local"}: exit status 1)
```

Also, note that it is just a **workarround** which we will able to remove when a stable release of the controller-runtime be released. So, we will do it properly. 

I really do not think that we should waste too much time on it and we can just apply the same solution which is working for Graphana and MSS Oper as was done [here](https://github.com/aerogear/app-metrics-operator/pull/11/files#diff-3661dc32b6bde1b0939ad409364e48fc) as well. 

We can just and copy the utills and utils_test.go and keep all equals which indeed in my POV will allow us to remove it horizontally and easier when the release with the feature/fix be available to us. 


**Also, this solution will not allow us to impl the unit-test until these releases be made and we upgrade the projects.**

pkg/nspredicate/nspredicate.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+// NewNamespacePredicateFromEnvVar looks up the provided environment
+// variable, parses the value, and returns a NamespacePredicate
+func NewFromEnvVar(envVar string) (NamespacePredicate, error) {
+	namespaces, found := os.LookupEnv(envVar)
+	if !found {
+		return NamespacePredicate{}, fmt.Errorf("Environment variable %s not found", envVar)
```
camilamacedo86: It is working with operator-sdk local and unit tests? 
This project had a unit test added and coveralls which is not appearing now in this PR as the CI validation. Do you know why?

pkg/controller/iosvariant/iosvariant_controller.go
grdryn: Hmm, which command are you running to see that error? The existing tests appear to work fine for me on this branch:

```
make test/unit 
Running tests:
CGO_ENABLED=1 go test -v -race -cover github.com/aerogear/unifiedpush-operator/pkg/controller/androidvariant
=== RUN   TestReconcileAndroidVariant_Reconcile
--- PASS: TestReconcileAndroidVariant_Reconcile (0.01s)
PASS
coverage: 16.1% of statements
ok  	github.com/aerogear/unifiedpush-operator/pkg/controller/androidvariant	1.157s	coverage: 16.1% of statements
```

camilamacedo86: It is working fine because in this test we are not checking the objects managed and created by it. It was added just as an example.  I can do a pr with a test over it if you wish. However, since it is really a workaround is valid to spend time on this solution? I will do a fast test just fr we check it anyway.

camilamacedo86: @grdryn let's merge the following PR then we can check if it will work or not properly. 

https://github.com/aerogear/unifiedpush-operator/pull/47

wei-lee (approved):

odra (approved):

grdryn:
Thanks for the reviews! I'm going to hold off on merging this until after @secondsun has gotten #31 through, since it would be easier to deal with any potential conflicts here than there (I doubt there would be any anyway).

I'll also create a PR to the installer repo to add the new env vars to the operator (thanks for spotting that @odra!).

jomrazek (approved):
working as expected

PR #50, opened 2019-07-18
Update UnifiedPush Server image to version 2.3.0

danielpassos:


pkg/config/config.go
```diff
@@ -42,7 +42,8 @@ func New() Config {
 		PostgresImageStreamTag: getEnv("POSTGRES_IMAGE_STREAM_TAG", "10"),
 
 		// these are used when the image stream does not exist and created for the first time by the operator
-		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-wildfly-plain:2.2.1.Final"),
+		// UPSImageStreamInitialImage:     getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "docker.io/aerogear/unifiedpush-configurable-container:2.3.0"),
+		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "quay.io/aerogear/unifiedpush-configurable-container:2.3.0"),
```
grdryn:
```suggestion
		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "quay.io/aerogear/unifiedpush-configurable-container:2.3"),
```

@danielpassos Myself and @wei-lee were just talking about this, and we think maybe using the `2.3` tag might be better. That way, patch fixes in the UPS project (e.g. `2.3.1`) can be released without needing a new release of the operator. What do you think?

(the process to update to the `2.3.1` release would just be that the SRE team would do an `oc import-image ups-imagestream` during their scheduled maintenance window)

Also, could you change the default for `UPS_IMAGE_STREAM_TAG` on line 34 to be `2.3` also?

danielpassos: Done!

danielpassos:
@grdryn I'm wondering if I should also change the `UPS_IMAGE_STREAM_NAME`

grdryn (approved):
@danielpassos I don't think it's necessary to change the image stream name one (I'm not sure if that's even a worthwhile thing to allow modification of).

This looks perfect :+1:

secondsun (approved):

PR #52, opened 2019-07-22
Tighten permissions in operator ClusterRole

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9616

## Verification Steps

- Check out the branch for this PR
- `make cluster/prepare`
- `kubectl apply -f deploy/operator.yaml`
- Check the operator pod logs to make sure there aren't any errors which could be related to permissions
- `kubectl apply -n unifiedpush -f deploy/crds/push_v1alpha1_unifiedpushserver_cr.yaml`
- Check the operator pod logs again to make sure there aren't any errors which could be related to permissions
- Wait for the UPS to become ready

- Create app CRs in the apps namespace, and check the operator pod logs again to make sure there aren't any errors which could be related to permissions.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -3,47 +3,105 @@ kind: ClusterRole
 metadata:
   name: unifiedpush-operator
 rules:
+
+# The resources in this group are the CRs that the controllers in this
+# operator react to. Nothing in the operator is expected to create or
+# delete CRs, so those permissions are not given.
+- apiGroups:
+  - push.aerogear.org
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+
+# For "secondary resources", the operator needs to be able to do
+# pretty much everything. The "deletecollection" permission is the
+# only one not given here (if it needs to delete more than one
+# instance of a kind, it can do them one-by-one.
 - apiGroups:
   - ""
   resources:
-  - pods
   - services
-  - endpoints
   - persistentvolumeclaims
   - events
   - configmaps
   - secrets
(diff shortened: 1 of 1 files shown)


deploy/role.yaml
```diff
@@ -3,47 +3,106 @@ kind: ClusterRole
...
 rules:
+
+# The resources in this group are the CRs that the controllers in this
+# operator react to. Nothing in the operator is expected to create or
+# delete CRs, so those permissions are not given.
+- apiGroups:
+  - push.aerogear.org
+  resources:
+  - '*'
+  verbs:
```
odra, line 14: are those all the verbs? if so can't you use  `*`?

odra: nvm, patch is not in the list :)

grdryn: Also delete is not there. Actually 'create' should maybe not be there either (see comment above this section, I intended to remove it)

aliok:
I didn't verify the changes but they look fine.

IMO, let's not spend too much time on this optimization. We can always do optimizations in a later stage.
We're giving the operator shitload of access to lots of stuff anyway, so e.g. removing 'create' access doesn't mean that much to me.
But I appreciate the changes here :)

odra (approved):

camilamacedo86:
Hi, @odra and @aliok these changes broke the Monitoring features. Also, it is something that affects all project and a full regression test should be performed in order to validate all features as if the backup will still working as well. IMHO: It is not really something that we should do now. 

HI @grdryn I will open a PR to revert it.

PR #57, opened 2019-07-31
sop improvements

laurafitzgerald:
## Motivation
I was working on SOPs for MDC as part of this JIRA: https://issues.jboss.org/browse/AEROGEAR-9116 and this JIRA: https://issues.jboss.org/browse/INTLY-1436
And spotted some improvements that could be made here as I was basing my changes on this SOP

## What
Language improvements

## Additional Notes
**NOTE** This is a draft pr as I'm pushing it to save it, but there could be further work to be done here. There is another piece of work which is priority over this one now.

### DIFF
--- a/SOP/SOP-push.adoc
+++ b/SOP/SOP-push.adoc
@@ -18,7 +18,7 @@ toc::[]
 
 == Overview
 
-The following guide outlines the steps required to manage and solve issues over the https://github.com/aerogear/aerogear-unifiedpush-server[UnifiedPush Server] managed, installed and configured via the https://github.com/aerogear/unifiedpush-operator[UnifiedPush Operator].
+The following guide outlines the steps required to manage and solve issues in the https://github.com/aerogear/aerogear-unifiedpush-server[UnifiedPush Server] which is managed, installed and configured via the https://github.com/aerogear/unifiedpush-operator[UnifiedPush Operator].
 
 == Reference Articles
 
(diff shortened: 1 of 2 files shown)


SOP/SOP-push.adoc
```diff
@@ -32,9 +32,7 @@ All alerts should appears as green in the Prometheus Alert Monitoring.
 
 == Prometheus Alerts Procedures
 
-The following guide outlines the steps required to solve issues which will be alert by the Prometheus alerts configured to monitoring the  https://github.com/aerogear/aerogear-unifiedpush-server[UnifiedPush Server] when it is managed, installed and configured via the https://github.com/aerogear/unifiedpush-operator[UnifiedPush Operator].
```
camilamacedo86, line 35: I do not think that it should be removed. It brings context and explains what the goal of it which was one of the requests made by @Nanyte25.

laurafitzgerald: This line is duplicated above.

camilamacedo86 (approved):
Well done !!

SOP/SOP-push.adoc
```diff
@@ -32,9 +32,7 @@ All alerts should appears as green in the Prometheus Alert Monitoring.
...
-
-IMPORTANT: Before check any following steps see if the operator pod is running successfully. More info: https://github.com/aerogear/unifiedpush-operator/blob/0.1.0/SOP/SOP-operator.adoc[UnifiedPushOperatorDown]
+IMPORTANT: Before checking any of following steps see if the operator pod is running successfully. More info: https://github.com/aerogear/unifiedpush-operator/blob/0.1.0/SOP/SOP-operator.adoc[UnifiedPushOperatorDown]
```
grdryn: Should/could this be a file link to that other document, since they live side-by-side? That could remove the need to reference the (outdated) version in the URL.

grdryn (approved):

SOP/SOP-push.adoc
camilamacedo86: The link is important for the SRE checked it. It is important to remeber that the SOPs should be made thinking in someone that never saw it before. So, besides, be side-side the person will not guess it n POV. Also, note that we have already planned to create a var for all <tags> so it will not be a problem at all. For now, a find/replace from IDE should help us to do it easily.

grdryn: I'm not suggesting that we remove the link, I'm just suggesting that it could be a local one rather than a remote one.

camilamacedo86: Ok. It is fine .. would be better .. but a really small adjustment.

laurafitzgerald: I've made the change as suggest. I've verified myself that it works but @grdryn could you verify it quickly and then I will merge this one. :)

grdryn: Verified! :100:

PR #60, opened 2019-08-06
 Make delete methods in client package idempotent

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9697

Among other things, this will allow variant CRs to be deleted if they
no longer exist in UPS (which can happen if the PushApplication is
deleted first).

## Verification Steps

Create a PushApplication CR and then an AndroidVariant CR underneath it (use the ID of the PushApplication in the variant CR). Once those are created, first delete the PushApplication CR, and then delete the AndroidVariant CR. The AndroidVariant CR should get deleted successfully, and there shouldn't be any errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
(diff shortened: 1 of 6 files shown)


ziccardi:
👀

laurafitzgerald:
looks like @ziccardi is reviewing. I will leave it with him. Ping me again if you want additional reviews/verification.

grdryn:
Bump!

ziccardi (approved):
Tested in my cluster 👍

PR #62, opened 2019-08-16
adding resource limits

secondsun:
## Motivation
Set some initial resource limits for ups and the operator.  From this JIRA : https://issues.jboss.org/browse/AEROGEAR-9765?

## What
More or less the limits I set were in line with the suggestions in the JIRA, where they deviate I had found other deployments of the same service that had limits set, or for ups server I tested with lower memory/cpu considerations

## Verification Steps
Standard operator smoke tests should be fine.

## Additional Notes

The limts are fluid right now and we are seeking feedback

### DIFF
--- a/pkg/controller/unifiedpushserver/unifiedpushserver.go
+++ b/pkg/controller/unifiedpushserver/unifiedpushserver.go
@@ -3,6 +3,8 @@ package unifiedpushserver
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	pushv1alpha1 "github.com/aerogear/unifiedpush-operator/pkg/apis/push/v1alpha1"
 	openshiftappsv1 "github.com/openshift/api/apps/v1"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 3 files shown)


camilamacedo86 (approved):
Should be good have the limits in the CR's as it is done for the operator, however, the other configurations are fixed in the code as well. So, it is fine for me.

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -315,6 +327,16 @@ func newUnifiedPushServerDeploymentConfig(cr *pushv1alpha1.UnifiedPushServer) (*
 									ContainerPort: 4180,
 								},
 							},
+							Resources: corev1.ResourceRequirements{
+								Limits: corev1.ResourceList{
+									"memory": resource.MustParse("2Gi"),
+									"cpu":    resource.MustParse("20m"),
+								},
+								Requests: corev1.ResourceList{
+									"memory": resource.MustParse("2Gi"),
```
wei-lee: do we need this much memory for the oauth proxy?

pkg/controller/unifiedpushserver/unifiedpushserver.go
```diff
@@ -270,6 +272,16 @@ func newUnifiedPushServerDeploymentConfig(cr *pushv1alpha1.UnifiedPushServer) (*
...
 							ImagePullPolicy: corev1.PullAlways,
 							Env:             buildEnv(cr),
+							Resources: corev1.ResourceRequirements{
+								Limits: corev1.ResourceList{
+									"memory": resource.MustParse("2Gi"),
+									"cpu":    resource.MustParse("1"),
+								},
+								Requests: corev1.ResourceList{
+									"memory": resource.MustParse("512Mi"),
+									"cpu":    resource.MustParse("100m"),
```
wei-lee: should we request a bit more CPU, maybe start with 500m?

pkg/controller/unifiedpushserver/unifiedpushserver.go
secondsun: Yeah looks like we updated the values in the spreadhseet.  I'll make it 32/64mi

pkg/controller/unifiedpushserver/unifiedpushserver.go
secondsun: Sure thing

secondsun:
@camilamacedo86 I'll make a new JIRA ticket for making the resource limits configurable in the UPS CR.

secondsun:
Added the JIRA here : https://issues.jboss.org/browse/AEROGEAR-9778

PR #67, opened 2019-08-26
AEROGEAR-9778 - Add resource limits to UPS CRs

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9778

I basically added parameters in the Push Server CR to replace the hardcoded values here: https://github.com/aerogear/unifiedpush-operator/pull/62

Verification:
1. Try `oc apply -f push_v1alpha1_unifiedpushserver_cr.yaml` and you should see the resource requests/limits on OpenShift web console just as described on https://issues.jboss.org/browse/AEROGEAR-9765
2. Try `oc apply -f push_v1alpha1_unifiedpushserver_cr_with_resource_limits.yaml` and you should see different values
3. Change the values in the UI and see them reconciled
4. Leave some values empty in the CR and some filled. The defaults should be used for the empty ones, but CR should not be touched.


secondsun:
Not marking this as a change request yet, but I'd like a bit of discussion.

Instead of making each parameter its own variable in the yaml, can we reuse the resource limit objects from k8s?

For example

```yaml
spec:
  unifiedPushMemoryLimit: "3Gi"
  unifiedPushMemoryRequest: "1Gi"
  unifiedPushCpuLimit: "2"
  unifiedPushCpuRequest: "1"
  oAuthMemoryLimit: "128Mi"
  oAuthMemoryRequest: "64Mi"
  oAuthCpuLimit: "100m"
  oAuthCpuRequest: "50m"
  postgresMemoryLimit: "2Gi"
  postgresMemoryRequest: "1Gi"
  postgresCpuLimit: "2"
  postgresCpuRequest: "1"
  postgresPVCSize: "10Gi"
```
becomes
```yaml
spec:
  upsResources:
    limits:
      memory: 512Mi
      cpu: 1
    requests:
      memory: 256Mi
      cpu 500m
  oauthProxyResources:
    limits:
      memory: 128Mi
    requests:
      memory: 64Mi
#etc
```

aliok:
@secondsun actually I also noticed ~~that~~ the flat structure but I used the convention from other operators. @camilamacedo86 what do you think?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -736,3 +739,63 @@ func containsCronJob(cronJobs []batchv1beta1.CronJob, candidate *batchv1beta1.Cr
 	}
 	return false
 }
+
+func applySpecDefaultValues(instance *pushv1alpha1.UnifiedPushServer) {
+
+	if instance.Spec.UnifiedPushMemoryLimit == "" {
+		instance.Spec.UnifiedPushMemoryLimit = cfg.UPSMemoryLimit
```
grdryn: Does this actually set the value in the CR spec? Is it a good idea for operators to update spec fields in CRs? :thinking:

grdryn (commented):
I like @secondsun's suggestion.

Also, does this actually do any reconciliation of the values? E.g. what happens if I update my CR spec to set a higher request or limit than I had originally set?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: similarly, I just took the example from our other operators.

aliok:
@grdryn 

> Also, does this actually do any reconciliation of the values? E.g. what happens if I update my CR spec to set a higher request or limit than I had originally set?

That's something I didn't think of. I don't think a reconciliation is done in that case ATM.

camilamacedo86:
Hi @secondsun and @aliok,

Regards the specs; if it is not flat then it means that the objects/structures to receive them need to be created which will require spending more effort here to do it. I am ok with all in string format, however, is up to you folks decide to spend effort on it.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Can you share a link to those?

camilamacedo86: I'd recommend do it as it was done here[1]. Create a config which will set the values when they are not defined.  

[1] - https://github.com/aerogear/mobile-security-service-operator/blob/master/pkg/controller/mobilesecurityservice/mandatory_specs.go

secondsun:
@camilamacedo86 If we use the k8s resource types then we don't need to create new types; we can just use the existing ones.  So instead of each service creating 4 new fields each service only adds one : the resourceReuest

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: https://github.com/aerogear/mobile-security-service-operator/blob/master/pkg/controller/mobilesecurityservice/controller.go#L175

There's 1 only. I thought there was more than one.

aliok:
@secondsun @grdryn @camilamacedo86 
Thanks for the suggestions all. I will do these:
* Use the resource struct type from K8s/OpenShift in CRD
* Reconcile the values

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
camilamacedo86: In MSS Oper we have a default setup for each type in order to not hurt concepts as cohesion and encapsulation.

secondsun:
@aliok <3 You are a gentleman and a scholar

aliok:
With the latest commit:
* Spec is untouched
* `corev1.ResourceRequirements` is used in the spec, instead of flat list of parameters
* Reconciliation is done for the values in spec (resource requests/limits for DCs and Postgres PVC size)

@grdryn I wrote the code blindly because I am having trouble with my Minishit installation.
I will try to test it during the rest of the day (1.5 hours remaining).
If I can't make it, feel free to take it over during my PTO, or to leave it to me when I am back.

UPDATE: tested stuff locally and pushed another commit. Should be good now. Adding verification instructions to the PR desc.

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -69,6 +69,15 @@ spec:
                 - backendSecretName
                 type: object
               type: array
+            oAuthResourceRequirements:
+              type: object
```
aliok, line 73: We need to check if there's a way to have the nested fields written here using some `kubebuilder` magic.

camilamacedo86: what do you mean? Would you like to verify it?

aliok: I asked this in RHMI and OperatorFW channels and received no answer. Let's merge w/o this improvement

aliok:
@laurafitzgerald can you review and verify this?

aliok:
cc @jomrazek @odra @wei-lee

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -69,6 +69,15 @@ spec:
                 - backendSecretName
                 type: object
               type: array
+            oAuthResourceRequirements:
```
laurafitzgerald, line 72: Do we need to bump any version of the crd now that there is a change to it? cc @grdryn

laurafitzgerald:
Going through the verification steps now.

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
grdryn: I don't think so, given that we're only introducing optional new stuff. The existing API will still work as expected -- nothing has changed or been removed there.

laurafitzgerald:
Verification is now completed

- [x] Try oc apply -f push_v1alpha1_unifiedpushserver_cr.yaml and you should see the resource requests/limits on OpenShift web console just as described on https://issues.jboss.org/browse/AEROGEAR-9765
- [x] Try oc apply -f push_v1alpha1_unifiedpushserver_cr_with_resource_limits.yaml and you should see different values
- [x] Change the values in the UI and see them reconciled
- [x] Leave some values empty in the CR and some filled. The defaults should be used for the empty ones, but CR should not be touched.

aliok:
Aw yeah!

PR #69, opened 2019-09-05
Use annotations for storing Ids; reconcile secrets

grdryn:
We previously relied on Ids (from UPS) stored in the Status block of
the CRs, but that led to problems when trying to restore from
backups.

This change doesn't remove the Ids from the Status block, since we
know the MDC relies on them being there for the moment. We should
consider updating that to get it from the annotation, and then no
longer present it in the Status block.

I've also updated some of the reconciliation logic to update the
secrets in the Status block, if someone has performed a "Renew Secret"
through the UPS Admin UI.

----

## Verification Steps

- Deploy this version of the operator to an OpenShift cluster & create a UPS CR (I currently have it deployed, so I can save you some time here)
- Create a PushApplication & verify that the PushApplicationId exists both in the Status block, and as an annotation in `metadata.annotations`
- Grab the PushApplicationId from that, and create Android and iOS Variants. Similarly to the previous step, verify that the VariantId for each is set in the Status block and as an annotation

- From the UPS Admin UI, renew both the master secret for the application, and the secrets on the variants, and notice that they will get updated in the CR Statuses. This may take a while, so I've done something like these commands in 3 different terminals, and just waited until the output changed:

```
while true; do oc get pushapplication example-pushapplication -n unifiedpush-apps -o jsonpath='{.status.masterSecret}' && echo && sleep 10; done
while true; do oc get androidvariant example-androidvariant -n unifiedpush-apps -o jsonpath='{.status.secret}' && echo && sleep 10; done
while true; do oc get iosvariant example-iosvariant -n unifiedpush-apps -o jsonpath='{.status.secret}' && echo && sleep 10; done
```

^while those are running, you could stop/start the operator if you wanted to force the update ;)

### DIFF
--- a/pkg/controller/androidvariant/androidvariant_controller.go
+++ b/pkg/controller/androidvariant/androidvariant_controller.go
@@ -124,32 +124,51 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	if foundVariant != "" {
-		// we don't do a full reconciliation (update push app on UPS server based on CR content) but we
-		// only do initial creation of push apps.
-		reqLogger.Info("Skip reconcile: Android Variant already exists in UPS", "AndroidVaraint.Name", instance.Name)
-		return reconcile.Result{}, nil
+	if foundVariant.VariantId == "" {
+		createdVariant, err := unifiedpushClient.CreateAndroidVariant(instance)
+		if err != nil {
+			reqLogger.Error(err, "Error creating Android variant in UPS.", "AndroidVariant.Name", instance.Name)
+			return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+		}
+
+		if instance.ObjectMeta.Annotations == nil {
+			instance.ObjectMeta.Annotations = make(map[string]string)
+		}
+		instance.ObjectMeta.Annotations["variantId"] = createdVariant.VariantId
+		err = r.client.Update(context.TODO(), instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		reqLogger.Info("Android Variant created", "Name", instance.Name, "VariantId", createdVariant.VariantId)
+		return reconcile.Result{Requeue: true}, nil
+
 	}
 
-	variantId, secret, err := unifiedpushClient.CreateAndroidVariant(instance)
-	if err != nil {
-		reqLogger.Error(err, "Error creating Android variant in UPS.", "AndroidVaraint.Name", instance.Name)
-		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+	if instance.ObjectMeta.Annotations["variantId"] != foundVariant.VariantId {
+		instance.ObjectMeta.Annotations["variantId"] = foundVariant.VariantId
+		err = r.client.Update(context.TODO(), instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
-	instance.Status.VariantId = variantId
-	instance.Status.Secret = secret
-	instance.Status.Ready = true
-	err = r.client.Status().Update(context.TODO(), instance)
-	if err != nil {
-		reqLogger.Error(err, "Error updating AndroidVariant status", "AndroidVariant.Name", instance.Name)
-		return reconcile.Result{}, err
+	// If either of these have diverged, fix both, so there's just a single status update
+	if instance.Status.VariantId != foundVariant.VariantId || instance.Status.Secret != foundVariant.Secret || instance.Status.Ready != true {
+		instance.Status.VariantId = foundVariant.VariantId
+		instance.Status.Secret = foundVariant.Secret
+		instance.Status.Ready = true
+		err = r.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			reqLogger.Error(err, "Error updating AndroidVariant status", "Name", instance.Name)
+			return reconcile.Result{}, err
+		}
 	}
 
 	if err := util.AddFinalizer(r.client, reqLogger, instance); err != nil {
 		return reconcile.Result{}, err
 	}
 
-	reqLogger.Info("Android Variant created", "AndroidVariant.Name", instance.Name)
+	reqLogger.Info("Android Variant reconciled", "Name", instance.Name, "VariantId", foundVariant.VariantId)
 	return reconcile.Result{}, nil
 }
--- a/pkg/controller/iosvariant/iosvariant_controller.go
+++ b/pkg/controller/iosvariant/iosvariant_controller.go
@@ -8,9 +8,7 @@ import (
 	"github.com/aerogear/unifiedpush-operator/pkg/controller/util"
 	"github.com/aerogear/unifiedpush-operator/pkg/nspredicate"
 
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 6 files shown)


pkg/controller/androidvariant/androidvariant_controller.go
```diff
@@ -124,32 +125,48 @@ func (r *ReconcileAndroidVariant) Reconcile(request reconcile.Request) (reconcil
...
-		reqLogger.Info("Skip reconcile: Android Variant already exists in UPS", "AndroidVaraint.Name", instance.Name)
-		return reconcile.Result{}, nil
+	if (foundVariant == unifiedpush.AndroidVariant{}) {
+		createdVariant, err := unifiedpushClient.CreateAndroidVariant(instance)
+		if err != nil {
+			reqLogger.Error(err, "Error creating Android variant in UPS.", "AndroidVaraint.Name", instance.Name)
+			return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+		}
+
+		instance.ObjectMeta.Annotations["VariantId"] = createdVariant.VariantId
```
grdryn:
```suggestion
		instance.ObjectMeta.Annotations["variantId"] = createdVariant.VariantId
```

I'll make this change & same for other annotations modified here.

laurafitzgerald:
Faced this error when creating a PushApplication
```
{"level":"info","ts":1567699127.3251884,"logger":"controller_pushapplication","msg":"Reconciling PushApplication","Request.Namespace":"unifiedpush-apps","Request.Name":"laura-example-pushapplication"}
E0905 15:58:47.424981       1 runtime.go:69] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
```
More info in the pod logs
@grdryn is investigating

pkg/controller/iosvariant/iosvariant_controller.go
```diff
@@ -126,55 +124,51 @@ func (r *ReconcileIOSVariant) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
-	if foundVariant != "" {
-		// we don't do a full reconciliation (update push app on UPS server based on CR content) but we
-		// only do initial creation of push apps.
-		reqLogger.Info("Skip reconcile: IOS Variant already exists in UPS", "IOSVaraint.Name", instance.Name)
-		return reconcile.Result{}, nil
+	if foundVariant.VariantId == "" {
```
wei-lee, line 127: in the android variant controller, we are comparing the object. But here we compare the just the `VariantId` field. Should we keep them consistent?

pkg/controller/iosvariant/iosvariant_controller.go
```diff
@@ -126,55 +124,51 @@ func (r *ReconcileIOSVariant) Reconcile(request reconcile.Request) (reconcile.Re
...
+	// If either of these have diverged, fix both, so there's just a single status update
+	if instance.Status.VariantId != foundVariant.VariantId || instance.Status.Secret != foundVariant.Secret || instance.Status.Ready != true {
+		instance.Status.VariantId = foundVariant.VariantId
+		instance.Status.Secret = foundVariant.Secret
+		instance.Status.Ready = true
+		err = r.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			reqLogger.Error(err, "Error updating IOSVariant status", "Name", instance.Name)
+			return reconcile.Result{}, err
+		}
```
wei-lee, line 165: probably not something we need to fix now, but it does look like there are quite a lot of duplicated code between the android and ios variant controller. Should be able to do something about them.

pkg/unifiedpush/client.go
```diff
@@ -54,30 +54,30 @@ type UnifiedpushClient struct {
...
 
 // GetApplication does a GET for a given PushApplication based on the PushApplicationId
-func (c UnifiedpushClient) GetApplication(p *pushv1alpha1.PushApplication) (string, error) {
-	if p.Status.PushApplicationId == "" {
+func (c UnifiedpushClient) GetApplication(p *pushv1alpha1.PushApplication) (PushApplication, error) {
+	if p.ObjectMeta.Annotations["pushApplicationId"] == "" {
 		// We haven't created it yet
-		return "", nil
+		return PushApplication{}, nil
 	}
```
wei-lee, line 68: does this actually backward compatible? Should we also check if the id is in the status field too? Just thinking about existing apps, they will only have id in the status field?

Another option would be to add the annotation if the application id only exists in the status field but not in annotations?

pkg/unifiedpush/client.go
```diff
@@ -134,32 +134,31 @@ func (c UnifiedpushClient) DeleteApplication(p *pushv1alpha1.PushApplication) er
...
 
 // GetAndroidVariant does a GET for a given Variant based on the VariantId
-func (c UnifiedpushClient) GetAndroidVariant(v *pushv1alpha1.AndroidVariant) (string, error) {
-	if v.Status.VariantId == "" {
+func (c UnifiedpushClient) GetAndroidVariant(v *pushv1alpha1.AndroidVariant) (AndroidVariant, error) {
+	if v.ObjectMeta.Annotations["variantId"] == "" {
 		// We haven't created it yet
-		return "", nil
+		return AndroidVariant{}, nil
 	}
```
wei-lee, line 155: same question here

pkg/controller/iosvariant/iosvariant_controller.go
grdryn: Yeah, I think when I created the controllers originally, I wanted to just have one "Variant" controller, but given that they two different types have mutually exclusive fields, I figured it could save some hassle to just duplicate the shared code for now (rather than trying to determine based on the fields which type it was, and making sure it had all necessary fields but no invalid fields). We can definitely try to share code between them if we find that it's easy and makes sense :+1:

pkg/controller/iosvariant/iosvariant_controller.go
grdryn: Yeah, good point. I think I wanted to compare the object here too, but for some reason in this case it wouldn't work. I'll try again, and if it doesn't work I'll change the others to be like this one.

*Edit:* it's because the `Certificate` field in the `IOSVariant` type from the unifiedpush package has the type `[]byte`:

```
pkg/controller/iosvariant/iosvariant_controller.go:128:5: invalid operation: foundVariant == unifiedpush.IOSVariant literal (struct containing []byte cannot be compared)
```

I'll update the AndroidVariant and PushApplication controllers to work the same way here, by checking their id fields.

pkg/unifiedpush/client.go
grdryn: I think the updated code here does what you are talking about.

grdryn:
@wei-lee I think I've addressed all of your comments here now. Could you take another look when you have a moment?

laurafitzgerald:
- [x] Create a PushApplication & verify that the PushApplicationId exists both in the Status block, and as an annotation in metadata.annotations

- [x] Grab the PushApplicationId from that, and create Android and iOS Variants. Similarly to the previous step, verify that the VariantId for each is set in the Status block and as an annotation

- [x] From the UPS Admin UI, renew both the master secret for the application, and the secrets on the variants, and notice that they will get updated in the CR Statuses.

laurafitzgerald (approved):
I didn't do an in depth code review as I see that Wei has done that already. Adding verification as I've completed all the verification steps in the description successfully.

PR #70, opened 2019-09-09
fix: sender id is set correctly now

secondsun:
This fixes https://issues.jboss.org/browse/AEROGEAR-9881

To test this fix create the push app and the android variant using the cr's in the example directory and then confirm that they are set correctly in the UPS console


wei-lee:
@secondsun Is this a big issue? Do we actually need this for the current 1.5 release?

secondsun:
No, but it was a quick fix that @jhellar raised.

jomrazek (approved):
Verified, working as expected :+1:

PR #71, opened 2019-09-18
WebPush support

secondsun:
## Motivation
WebPush support in UPS requires additional queues for sending messages.  This PR creates those queues.

## What
I've added two new queues to the list for the operator to manager, `WebPushMessageQueue` and `WebTokenBatchQueue`.  

## Verification Steps
 
1. Run the operator and deploy UPS normally
2. Verify the new queues are created.


secondsun:
@grdryn Not an immediate concern, but UPS will need extra queues added when we deploy web push.  Right now the operator will not check those queues if UPS is deployed.  There is probably some logic as well for how we want to upgrade the pod that should be added, so we can discuss that wherever you want.

Web Push in UPS isn't done and won't be in the immediate (ie this week) future.  So we have some time to think about this.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -2,10 +2,11 @@ package unifiedpushserver
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"reflect"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/resource"
```
grdryn, line 8: There are other packages from k8s.io/apimachinery down around line 27 -- maybe this could be put down there?

grdryn:
> There is probably some logic as well for how we want to upgrade the pod that should be added, so we can discuss that wherever you want.

I'm not sure what you mean here. Is there a new container to be added to the pod, or something else?

secondsun:
If someone has a running UPS instance and updates the operator, the current logic won't run the reconciler because it checks if UPS is running and I don't think it checks the queues it uses.

Here's the return in question : https://github.com/aerogear/unifiedpush-operator/blob/master/pkg/controller/unifiedpushserver/unifiedpushserver_controller.go#L237

grdryn:
@secondsun I believe that should only return if you've got more than one UPS instance, which we should never have, right?

secondsun:
@grdryn When https://github.com/aerogear/aerogear-unifiedpush-server/pull/1038 is merged and a new image is pushed to quay, this PR will be ready to be tested.  The only step left will be updating the UPS image version.

README.adoc
```diff
@@ -199,6 +199,46 @@ Here are all of the configurable fields in an IOSVariant:
...
+|The alias to be used to identify this server.  Must be a url or mailto link
+
+|privateKey
+|The VAPID private key
+
+|publicKey
+|The VAPID public key
+|===
+
+* Apply an WebPushVariantCR based on the example a link:./deploy/crds/examples/push_v1alpha1_webpushvariant_cr.yaml[WebPushVariant CR]  as follows:
```
grdryn:
```suggestion
* Apply a WebPushVariantCR based on the example link:./deploy/crds/examples/push_v1alpha1_webpushvariant_cr.yaml[WebPushVariant CR] as follows:
```

cmd/manager/main.go
```diff
@@ -173,18 +173,20 @@ func main() {
 		log.Info("Could not create metrics Service", "error", err.Error())
 	}
 
-	err = addMonitoringKeyLabelToService(cfg, operatorNamespace, service)
-	if err != nil {
-		log.Error(err, "Could not add monitoring-key label to operator metrics Service")
-	}
+	if service != nil {
```
grdryn, line 176: I guess without this there's a problem when running locally?

deploy/role.yaml
```diff
@@ -1,26 +1,19 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  creationTimestamp: null
   name: unifiedpush-operator
 rules:
-
-# The resources in this group are the CRs that the controllers in this
```
grdryn, line 7: Did your editor (or the operator-sdk cli remove these comments? Could you put them back for now?

pkg/apis/push/v1alpha1/webpushvariant_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// to sign messages and identify them as being from this server by the
+	// client.
+	PrivateKey string `json:"privateKey"`
+
+	// PublicKey is the public part of a VAPID keypair. This will
+	// be transmitted to the user in the mobile-services.json file and the client
+	// uses it to verify messages were send by this server.
+	PublicKey string `json:"publicKey"`
+
+	// Alias is send to a web push push service and must be either
```
grdryn:
```suggestion
	// Alias is sent to a web push push service and must be either
```

pkg/unifiedpush/client.go
```diff
@@ -335,6 +344,97 @@ func (c UnifiedpushClient) DeleteIOSVariant(v *pushv1alpha1.IOSVariant) error {
...
+	}
+
+	if id == "" {
+		// We haven't created it yet
+		return WebPushVariant{}, nil
+	}
+
+	url := fmt.Sprintf("%s/rest/applications/%s/webpush/%s", c.Url, v.Spec.PushApplicationId, id)
+
+	req, err := http.NewRequest(http.MethodGet, url, nil)
```
grdryn, line 363: Is it odd that we don't check the error on this? :thinking: I see we don't do it on the other methods in here either, so don't feel like you need to change it here now.

secondsun:
I've rebased this image off https://github.com/aerogear/unifiedpush-operator/pull/77; it includes the updates to the ups version now.

secondsun:
@grdryn  Just letting you know this is really ready for review now

README.adoc
```diff
@@ -199,6 +199,46 @@ Here are all of the configurable fields in an IOSVariant:
...
+|ID of the PushApplication that this variant corresponds to
+
+|description
+|Human friendly description for the variant
+
+|alias
+|The alias to be used to identify this server.  Must be a url or mailto link
+
+|privateKey
+|The VAPID private key
```
grdryn, line 232: What's a VAPID?

cmd/manager/main.go
secondsun: Yeah, the monitor service isn't started / installed by default locally but the operator was expecting it and crashing.

README.adoc
secondsun: VAPID is the key format that webpush uses.  It is self registering which means that we don't need to do complicated setup with webpush servers (FCM, WNS, and autopush are all webpush servers that support vapid).

There are online tools to generate keys : https://tools.reactpwa.com/vapid as well as command line tools.

In fact we COULD make the keys auto generate, but I think that it makes more sense to provide them since they will probably be reused by developers during the development cycle and chosen/tested/trusted by ops during deployment.

deploy/role.yaml
secondsun: I think it was the operator-sdk cli.  I'll put them back

pkg/controller/webpushvariant/webpushvariant_controller.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Reconcile reads that state of the cluster for a WebPushVariant
+// object and makes changes based on the state read and what is in the
+// WebPushVariant.Spec
+// Note:
+// The Controller will requeue the Request to be processed again if
+// the returned error is non-nil or Result.Requeue is true, otherwise
+// upon completion it will remove the work from the queue.
+func (r *ReconcileWebPushVariant) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
laurafitzgerald, line 77: In MSS we implemented some stuff to check the namespace that a consumption cr is found in. In this case we'd expect to action only crs created in MDC namespace. Do we need to do that here? Did we do it elsewhere in UPS? It might be moot because of permissions stuff changing but mentioning it just in case.

pkg/unifiedpush/client.go
```diff
@@ -31,6 +31,15 @@ type AndroidVariant struct {
 	variant
 }
 
+// webPushVariant is an internal struct used for convenient JSON
+// unmarshalling of the response received from UPS
+type WebPushVariant struct {
+	PrivateKey string
+	PublicKey  string
+	Alias      string
+	variant
```
laurafitzgerald, line 40: This looks odd to me.

pkg/unifiedpush/client.go
```diff
@@ -335,6 +344,97 @@ func (c UnifiedpushClient) DeleteIOSVariant(v *pushv1alpha1.IOSVariant) error {
 	return nil
 }
 
+// GetWebPushVariant does a GET for a given Variant based on the VariantId
+func (c UnifiedpushClient) GetWebPushVariant(v *pushv1alpha1.WebPushVariant) (WebPushVariant, error) {
+	id := ""
```
laurafitzgerald, line 349: Can you say `id string` here instead?

laurafitzgerald (commented):
Just leaving a comment review as I don't have specific requests just questions. I think there might be other changes coming too. Feel free to ping me again for another review.

pkg/controller/webpushvariant/webpushvariant_controller.go
grdryn: Yes, we do that here, but the Reconcile is not really the best place for it: rather than calling this method for CRs in all namespaces and returning early if they're not in the approved set, we set a predicate on the watch so that it will only enqueue events that are from namespaces in that set. See lines 43 to 55 for where that's being done (and let me know if you want me to explain any of that :+1: ).

pkg/unifiedpush/client.go
laurafitzgerald: @grdryn explained this one to me. Apparently it's a normal thing in go. Resolving this comment as nothing needs to be done.

pkg/controller/webpushvariant/webpushvariant_controller.go
laurafitzgerald: Hey, that makes sense, I wonder if we can implement that in MSS, wdyt?

grdryn: I don't see why not :)

laurafitzgerald: Ok. I'll create a JIRA for it. It's probably low priority so I'll mark it as such. Will resolve this comment now.

pkg/unifiedpush/client.go
grdryn: I think it would have to be as-is, or `var id string`, not `id string` or `id string := ""`

I'm fine with it as-is, I think it's probably more common for declaration & initialization of variables inside a function.

README.adoc
```diff
@@ -199,6 +199,46 @@ Here are all of the configurable fields in an IOSVariant:
...
+|Field Name |Description
+
+|pushApplicationId
+|ID of the PushApplication that this variant corresponds to
+
+|description
+|Human friendly description for the variant
+
+|alias
+|The alias to be used to identify this server.  Must be a url or mailto link
```
grdryn, line 229:
```suggestion
|The alias to be used to identify the sender.  Must be a url or mailto link
```

^I presume (based on the slightly longer description in the CRD).

deploy/role.yaml
grdryn: These comments appear to still be removed.

deploy/role.yaml
```diff
@@ -1,26 +1,19 @@
...
-# operator react to. Nothing in the operator is expected to create or
-# delete CRs, so those permissions are not given.
 - apiGroups:
   - push.aerogear.org
   resources:
   - '*'
+  - webpushvariants
```
grdryn, line 11: Is this already covered by the previous line?

I guess the operator-sdk now lists them explicitly, rather than using the wildcard.

grdryn (approved):
I have reviewed this and am happy with it. I haven't done any verification though.

laurafitzgerald:
I started to verify this one following the steps.
I installed the operator from a image build from the branch
Created the WebPush cr from the example Provided.
I faced the following error in the operator
```
"Error creating WebPush variant in UPS.","Request.Namespace":"unifiedpush-apps","Request.Name":"example-webpushvariant","WebPushVariant.Name":"example-webpushvariant","error":"UPS responded with status code: 404, but expected 201"
```

secondsun:
@laurafitzgerald Can you post your webpush CR?  I just tested it locally and it works, and I've tested it on RHPDS as well

laurafitzgerald:
Hi @secondsun 
I did some investigation, the cause of the error is due to the fact that the pushapplicationid in the sample cr file doesn't match any Push Application. I was running an old version of MDC which didn't have the web option so I've been creating the WebPush cr on the command line. To work around this I - 
1) Created a PushApplication using the cr in the example folder.
2) Captured the UID of the application from the Push UI URL.
3) Update that value in the WebPush cr in the example folder
4) Created that cr.

The 404 error is resolved now.

I am seeing a new error
```
Received HTTP status 400 with message 400 Bad Request \t keypairValid:the provided private key does not match with the public key
```
but I believe this probably caused by the fact I'm not using MDC also.

Could we make a note in the README that it's not recommended to use the example crs (or that the they should be used in certain order and values need to be update) and that UPS can be deployed using the operator but for use it should be interacted with via MDC bindings (as this manages the crs correctly) or via the Push Admin UI if being used by a community member who wants standalone UPS.

grdryn:
Thoughts:

> Could we make a note in the README that it's not recommended to use the example crs (or that the they should be used in certain order and values need to be update)

That sounds fine (or just put comments to that effect into the yaml files themselves even).

> and that UPS can be deployed using the operator but for use it should be interacted with via MDC bindings (as this manages the crs correctly)

There's no harm in mentioning the MDC, but I think _should_ is a bit strong -- _could/can_ would be more appropriate, in my opinion. Even though our primary target is now integreatly, where they've decided that they're going to run everything all the time, I kind of liked the original goal of mobile services which was "just deploy the services you want". Now maybe that involves _always_ running MDC, but there's no need to couple so tightly -- someone else might come up with novelty ways to use these resources (CLI?), but saying that they should use MDC might discourage that.

> or via the Push Admin UI if being used by a community member who wants standalone UPS.

So if using the operator without using MDC, don't bother with the PushApplication & Variant CRs at all? I guess that's one approach, but I think it would be odd to have docs to discourage usage of 4 out of 5 of the CRDs that this operator manages.

secondsun:
I'm fine with beefing up the README so it serves as a better jumping off point for what the prerequisites are for the different variants.  

Part of it could also be moving/renaming the "examples" to "templates".  That way people have a hint that they shouldn't be used as is.  WDYT?

laurafitzgerald:
@secondsun sounds good to me. :)

laurafitzgerald (approved):
Reverified. When I actually followed the steps it worked great! 🙈 The variant is created. Merge!

PR #72, opened 2019-09-18
Upgrade to operator sdk v0.10.0

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9913

- Upgrade to the latest version of the operator-sdk & dependencies
- Expose the CR Metrics on port 8686
- Create ServiceMonitor objects if that CRD is available on the cluster

Verification Steps:

Deploy the operator. Everything should work as expected, except now the Service should have 2 ports, and also a ServiceMonitor object should get created (if the ServiceMonitor CRD is on the cluster).


Gopkg.toml
```diff
@@ -69,15 +56,4 @@ required = [
...
-    non-go = false
-
-  [[prune.project]]
-    name = "k8s.io/gengo"
-    non-go = false
-  
-  [[prune.project]]
-    name = "github.com/enmasseproject/enmasse"
-    unused-packages = true
+  unused-packages = true
```
grdryn, line 59: This reduces the number of files that we have in `vendor/`, but it means that we have to run `dep ensure` again if we want to use any package from a current dependency that we're not currently using.

cmd/manager/main.go
```diff
@@ -155,3 +196,67 @@ func main() {
...
+	err = kclient.Update(context.TODO(), service)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// createServiceMonitor is a temporary fix until the version in the
+// operator-sdk is fixed to have the correct Path set on the Endpoints
```
grdryn, line 221: I'll create a separate PR for the operator-sdk to get the path set as expected.

jomrazek (approved):
Everything seems to work normally, service has 2 ports and also ServiceMonitor object was created.

aliok:
@grdryn 

> Expose the CR Metrics on port 8686
> Create ServiceMonitor objects if that CRD is available on the cluster

Are these coming with operator SDK upgrade automatically? Or were these required by the task?

grdryn:
@aliok They don't come automatically, but they're part of the new template with the new version of the sdk, so I updated that along with the version of the library.

PR #73, opened 2019-09-26
AEROGEAR-9966 - Reconcile image references in operator and convert DC…

aliok:
Delete things done during https://github.com/aerogear/unifiedpush-operator/pull/28

And reapply things done during https://github.com/aerogear/unifiedpush-operator/pull/8
Verification:
* Prepare the cluster, start the operator, install the CR --> UPS is running
* Change the image in one of the deployments
* You will see the image is reverted to what's defined in the operator

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,19 +7,9 @@ type Config struct {
 	PostgresContainerName   string
 	OauthProxyContainerName string
 
-	UPSImageStreamName string
-	UPSImageStreamTag  string
-
-	OauthProxyImageStreamName string
-	OauthProxyImageStreamTag  string
-
-	PostgresImageStreamNamespace string
-	PostgresImageStreamName      string
-	PostgresImageStreamTag       string
-
-	UPSImageStreamInitialImage        string
-	PostgresImageStreamInitialImage   string
-	OauthProxyImageStreamInitialImage string
+	UPSImage        string
+	PostgresImage   string
+	OauthProxyImage string
 
 	BackupImage string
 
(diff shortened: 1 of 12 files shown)


psturc (approved):
I'll leave the code review on somebody else, but the functionality seems to be fine, just verified that locally.

grdryn:
:eyes:

grdryn:
While this may work for a clean installation, it doesn't seem to work for me for the case that I care most about: I've got an existing UPS being managed by the operator from master, and I want to upgrade to this version of the operator. Everything goes haywire :-/

aliok:
@grdryn I just gave it a try, worked fine for me.
What I did is
* Installed the operator, created a UPS instance
* Changed the `UPS_IMAGE` env var for the operator and deployed it again, so that it enforces another image
* New image is used for UPS and everything went up an running.

How did you try it?

aliok:
@grdryn 
> I've got an existing UPS being managed by the operator from master, and I want to upgrade to this version of the operator.

Oh, now I get it...
I actually don't know to how much extent we should care about this.
Because, we won't be using image streams in the new operator. So, then, we should have a clean up logic in our code that deletes the unused things (image stream currently; DC in the future) that were used previously. Let me do some reading/experimenting about this.

aliok:
@grdryn we can delete the unused stuff with Ansible in case of RHMI.
we need to have a decision around this.
Option A:
* Migration logic in the code to delete imagestreams (we also keep the role for imagestreams)
* Patch the operator role in Ansible to make sure we have permissions to new stuff (deployment, etc.)

Option B:
* No migration logic in the code. Simply assume in the new version of the operator, that imagestreams are deleted magically and role is updated
* Have the migration code in Ansible

pkg/config/config.go
```diff
@@ -46,20 +36,9 @@ func New() Config {
...
-
-		PostgresImageStreamNamespace: getEnv("POSTGRES_IMAGE_STREAM_NAMESPACE", "openshift"),
-		PostgresImageStreamName:      getEnv("POSTGRES_IMAGE_STREAM_NAME", "postgresql"),
-		// Used both to set the tag, and also for the "POSTGRES_VERSION" in the Secret
-		PostgresImageStreamTag: getEnv("POSTGRES_IMAGE_STREAM_TAG", "10"),
-
-		// these are used when the image stream does not exist and created for the first time by the operator
-		UPSImageStreamInitialImage:        getEnv("UPS_IMAGE_STREAM_INITIAL_IMAGE", "quay.io/aerogear/unifiedpush-configurable-container:2.3"),
-		OauthProxyImageStreamInitialImage: getEnv("OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE", "quay.io/openshift/origin-oauth-proxy:4.2.0"),
+		UPSImage:        getEnv("UPS_IMAGE", "quay.io/aerogear/unifiedpush-configurable-container:2.3"),
```
laurafitzgerald: Should we have UPS_IMAGE? Based on this `3. A version of the operator should handle one version of the operand (the service version).` my understanding is that we don't want to make this configurable. Updating the operator version should be the path to upgrade the service right?

aliok: Got rid of them now

pkg/controller/unifiedpushserver/postgresql.go
```diff
@@ -153,15 +139,15 @@ func newPostgresqlDeploymentConfig(cr *pushv1alpha1.UnifiedPushServer) (*openshi
 							},
 							Resources: getPostgresResourceRequirements(cr),
 							VolumeMounts: []corev1.VolumeMount{
-								corev1.VolumeMount{
+								{
```
grdryn, line 142: Interesting :thinking: Is the type not needed in these cases?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -482,6 +473,26 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 				}
 			}
 		}
+
+		desiredImage := constants.PostgresImage
+
+		containerSpec := findContainerSpec(foundPostgresqlDeploymentConfig, cfg.PostgresContainerName)
+		if containerSpec == nil {
+			reqLogger.Info("Skipping image reconcile: Unable to find container spec in deployment", "Deployment.Namespace", foundPostgresqlDeploymentConfig.Namespace, "Deployment.Name", foundPostgresqlDeploymentConfig.Name, "ContainerSpec", cfg.PostgresContainerName)
```
grdryn, line 481: Out of curiosity, under what circumstances would we expect this to be `nil`?

grdryn:
@aliok Option A is my preferred route. We'll need to update the role for the operator ServiceAccount anyway to give it access to monitoring resources anyway (and Deployment in #74).

The operator should do everything needed for the upgrade, including cleaning up stuff that it previously created but which is no longer used. The fact that we've relied on the integreatly ansible installer for some stuff in the previous release is not necessarily good, and in some cases we did it because it was quicker.

That said, with the general case aside (operator should do everything), if it's going to be a lot of extra work in this case then we can re-evaluate other options, including:

1. Don't bother merging this until we branch for integreatly on openshift 4 (it won't be needed there, since we won't be upgrading. Something will be needed somewhere if we include it in a future release of integreatly on openshift 3 though)

2. Do it in the integreatly ansible installation just this once. I had a good argument for this, but I can't remember it now. I'll update this if I can remember it.

Both of those options are only thinking about integreatly though, and we should really be thinking of this operator independently of that IMHO.

pkg/controller/unifiedpushserver/postgresql.go
aliok: Yep, no needed.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
aliok: When there's e.g. no Postgres container definition within the Deployment's pod template

aliok:
As we talked recently, I will close this PR and merge it within https://github.com/aerogear/unifiedpush-operator/pull/74

PR #75, opened 2019-09-26
INTLY-3288 - Update Oauth Proxy Image

KevFan:
## Motivation
As part of [INTLY-3288](https://issues.jboss.org/browse/INTLY-3288), when trying to install the UPS operator and verifying it works on a Openshift 4 cluster, an `internal 500 error` was noticed when trying to login via the exposed route. However,  this does not happen on a Openshift 3 cluster.

The error is similar to https://bugzilla.redhat.com/show_bug.cgi?id=1678645 

## What
To resolve this, the `oauth-proxy` image was updated to a later version that supported login through Openshift 4 

## Why
This was done so that the user can successfully login via the exposed UPS route on a Openshift 4 cluster

## How
The image for `oauth-poxy` was update to point to `quay.io/openshift/origin-oauth-proxy:4.2.0`

## Verification Steps
* Using a openshift 4 cluster. install UPS operator using this branch. 
  * `oc login` to cluster
  * `make cluster/prepare`
  * `make code/run`
* After resources have been provisioned, verify that the user can login to UPS via the exposed route without errors
* (Optional Step) - To see error in login, the image tag in the imagestream can be reverted to point back to `docker.io/openshift/oauth-proxy:v1.1.0`
* Repeat for an openshift 3 cluster, to ensure compatability is not broken

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -409,7 +409,7 @@ environment variable names, along with their default values:
 | Tag of the Oauth proxy image stream that will be created by the operator.
 
 |`OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE`
-|`docker.io/openshift/oauth-proxy:v1.1.0`
+|`quay.io/openshift/origin-oauth-proxy:4.2.0`
 | Initial image for the Oauth proxy image stream that will be created by the operator.
 
 |`POSTGRES_IMAGE_STREAM_NAMESPACE`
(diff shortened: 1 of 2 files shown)


grdryn (approved):
This looks fine to me, and works on OpenShift 3 also.

Upgrading from an existing version of the operator won't actually cause the existing UPS install to be upgraded to this new image though -- someone would manually need to update the ImageStream. Don't worry about that though, as @aliok is removing imagestreams altogether in #73.

PR #80, opened 2019-11-06
Ups apns token

secondsun:
## Motivation
Implements support for IOSTokenVariant CRD which lets users setup a token based ios variant in UPS.

## What
 * Create iOSTokenVariant CRD, example CR and controller
 * Support Create, Get, and Delete operations for iOSTokenVariants
 * Validation of fields.

## Why
Tokens are an alternative way to communicate with APNS and is a feature the community has asked for.

## How
Originally, we wanted to overload the iOSVariant CR.  This proved to not be vaiable because the validation rules we wanted to set were not supported by code generation.  iOSTokenVariant was created in place of that; in the future we should probably refactor a bunch of this to use more regular patterns.

## Verification Steps
1. Deploy the operator with minishift
2. Deploy a ups and push application with the example crs.
3. Get a teamId, keyId, bundleId, and private key from APNS or from @ziccardi 
4. Edit deploy/crds/examples/push_v1alpha1_iostokenvariant_cr.yaml with those values
5. Confirm that Creating, reading, and deleting the variant records those actions in UPS correctly
6. If you have access to a iOS device you can test sending in UPS, but this has been testing in UPS already.

## Checklist:

- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task


grdryn:
:eyes:

pkg/constants/constants.go
```diff
@@ -1,7 +1,7 @@
 package constants
 
 const (
-	UPSImage        = "quay.io/aerogear/unifiedpush-configurable-container:2.4"
+	UPSImage        = "quay.io/aerogear/unifiedpush-configurable-container:2.5"
```
grdryn: Could you make this `2.5.0-1`? We don't want the floating tags any longer, and getting a different version of the operand should be done by getting the corresponding version of the operator.

pkg/controller/iostokenvariant/iostokenvariant_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		reqLogger.Error(err, "Error getting a UPS Client.", "IOSTokenVariant.Name", instance.Name)
+		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
+	}
+
+	// Check if the CR was marked to be deleted
+	if instance.GetDeletionTimestamp() != nil {
+		// First delete from UPS
+		err := unifiedpushClient.DeleteIOSTokenVariant(instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to delete IOSTokenVariant from UPS", "IOSTokenVaraint.Name", instance.Name)
```
grdryn:
```suggestion
			reqLogger.Error(err, "Failed to delete IOSTokenVariant from UPS", "IOSTokenVariant.Name", instance.Name)
```

deploy/olm-catalog/unifiedpush-operator/0.2.0/unifiedpush-operator.v0.2.0.clusterserviceversion.yaml
```diff
@@ -25,6 +25,11 @@ spec:
       version: v1alpha1
       displayName: Android Variant
       description: Represents a UPS Android Variant
+    - kind: IOSTokenVariant
+      name: iostokenvariants.push.aerogear.org
+      version: v1alpha1
+      displayName: iOS Variant
+      description: Represents a UPS iOS Variant
```
grdryn: Remove this, it's the CSV for an older version. When we release a new version this variant should then show up.

grdryn (approved):
Verified, looks good! :+1: 

I've made some very minor comments inline.

secondsun:
@grdryn fixed!

PR #81, opened 2019-11-08
✅ Fix broken reconciliation unit test

grdryn:
## Motivation
Want working unit tests

## What
Fixed the broken unit test

## Why
It was broken

## How
See "Files changed" tab

## Verification Steps
Testing done by Jenkins pipeline should be sufficient, I don't think there's any manual verification needed

## Checklist:

## Progress

- [x] Finished task
- [x] THIS TEMPLATE IS A NUISANCE


pkg/controller/unifiedpushserver/util.go
```diff
@@ -63,19 +64,27 @@ func updateContainerSpecImage(deployment *appsv1.Deployment, name string, image
...
-	apiLists, err := dc.ServerResources()
-	if err != nil {
-		return false, err
-	}
-	for _, apiList := range apiLists {
-		if apiList.GroupVersion == apiGroupVersion {
-			return true, nil
+// apiVersionChecker is just a container to make it easier to fake the
+// check function for tests
+type apiVersionChecker struct {
```
grdryn: Maybe something with an interface would be better here? I don't know how that would even look though

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -52,7 +53,17 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileUnifiedPushServer{client: mgr.GetClient(), config: mgr.GetConfig(), scheme: mgr.GetScheme()}
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		log.Error(err, "Failed to get clientset")
+		os.Exit(1)
```
grdryn, line 59: I don't know what to do here for this error case. Is it okay to completely bail like this?

This shouldn't ever be hit in any case, and if it is, I don't know how to continue, so exiting seems fair to me since it's still kind of at an initialization stage.

pkg/controller/unifiedpushserver/util.go
laurafitzgerald: Should it be in its own file if it's a struct with an interface?

grdryn: I've put it into its own file now :+1:

laurafitzgerald (approved):
@grdryn looks good and makes sense after the discussion we had. No verification as you've outline the passing tests in the pipeline are enough.

PR #83, opened 2019-11-13
operator creates monitoring resources for ups application

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-10055

## What
What was done in [this pr](https://github.com/aerogear/mobile-developer-console-operator/pull/58) except for the UPS operator

## Verification Steps
TODO 

## Checklist:

## Image
quay.io/aerogear/unifiedpush-operator:011d3d9

ping @grdryn for early code review if you have some time. It still needs to be deployed so the Grafana DashBoard can be checked and using the CreateOrUpdate and Reconcile syntax, it's a little more complex here than in the mdc operator resource pr. I'll push that in a seperate commit later today to ensure it's reviewable and correct.


laurafitzgerald:
@grdryn this is ready for verification now. It's deployed to the cluster/project I sent you earlier.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+- Operator creates the Monitoring Resources (GrafanaDashboard, PrometheusRule and ServiceMonitor) on creation of the UnifiedPushServer CR.
+- Operator creates its own Monitoring Resrouces (GrafanaDashboard, PrometheusRule) on installation.
```
grdryn, line 10:
```suggestion
- Operator creates its own Monitoring Resources (GrafanaDashboard, PrometheusRule) on installation.
```

:trollface:

Makefile
```diff
@@ -87,9 +87,6 @@ monitoring/install:
 	- kubectl apply -n $(NAMESPACE) -f deploy/monitor/operator_service.yaml
 	- kubectl apply -n $(NAMESPACE) -f deploy/monitor/prometheus_rule.yaml
 	- kubectl apply -n $(NAMESPACE) -f deploy/monitor/grafana_dashboard.yaml
-	- kubectl apply -n $(NAMESPACE) -f deploy/monitor/push_service_monitor.yaml
```
grdryn, line 90: I think this entire target (and the `monitoring/uninstall` one) can be removed now, right?

README.adoc
```diff
@@ -444,28 +444,13 @@ If you would like to modify the container names, you can use the following envir
 
 The application-monitoring stack provisioned by the
 {application_monitoring_operator} on https://github.com/integr8ly[Integr8ly]
-can be used to gather metrics from this operator and the UnifiedPush Server. These metrics can be used by Integr8ly's application monitoring to generate Prometheus metrics, AlertManager alerts and a Grafana dashboard.
+can be used to gather metrics from this operator and the UnifiedPush Server. These metrics can be used by Integr8ly's application monitoring to generate Prometheus metrics, AlertManager alerts and a dashboard in Grafana.
 
 It is required that the https://github.com/integr8ly/grafana-operator[integr8ly/Grafana] and https://github.com/coreos/prometheus-operator[Prometheus] operators are installed. For further detail see https://github.com/integr8ly/application-monitoring-operator[integr8ly/application-monitoring-operator].
 
-The following command enables the monitoring service in the operator namespace:
+The operator will install it's own monitoring resrouces required by Grafana and Prometheus on startup and will install the Resources required for monitoring the UnifiedPush Server on creation of the UnifiedPushServer CR.
```
grdryn, line 451:
```suggestion
The operator will install its own monitoring resources required by Grafana and Prometheus on startup and will install the Resources required for monitoring the UnifiedPush Server on creation of the UnifiedPushServer CR.
```

cmd/manager/main.go
```diff
@@ -146,15 +151,28 @@ func main() {
...
-	operatorNamespace, err := k8sutil.GetOperatorNamespace()
-	if err != nil {
+	// Setup Scheme for Integreatly Grafana apis
+	if err := integreatlyv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	// Setup Scheme for Prometheus Monitoring apis
+	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
```
grdryn, line 167: This looks like it's already being added on L155 -- I guess this is duplicated since you had it before you included my PR.

cmd/manager/main.go
```diff
@@ -188,55 +224,36 @@ func main() {
...
+	client := mgr.GetClient()
+	prometheusRule := &monitoringv1.PrometheusRule{ObjectMeta: metav1.ObjectMeta{Name: "unifiedpush-operator", Namespace: operatorNamespace}}
 
-	err = kclient.Update(context.TODO(), service)
-	if err != nil {
-		return err
-	}
+	controllerutil.CreateOrUpdate(ctx, client, prometheusRule, func(ignore k8sruntime.Object) error {
+		reconcilePrometheusRule(prometheusRule)
+		return nil
```
grdryn, line 240: The PrometheusRule & GrafanaDashboard created here don't have any owner set, so removing the operator won't cause them to be removed. Maybe the ServiceMonitor should be set as the owner (if that gets deleted, these other resources are of no use)?

That would also mean that creation/updation of these should happen inside the `if service != nil {}` on line 194, but maybe that's not a bad thing either. WDYT?
*Edit:* actually it will need to be put inside that condition anyway so that we can still run the operator locally (not in cluster).

If you're adding an owner reference, I'd do it here, between `reconcilePrometheurRule(prometheusRule)` and `return nil`.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -939,6 +978,50 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 	}
 	//#endregion
 
+	//#region Monitoring
+	//## region ServiceMonitor
+	serviceMonitor := &monitoringv1.ServiceMonitor{ObjectMeta: metav1.ObjectMeta{Name: "unifiedpush", Namespace: instance.Namespace}}
```
grdryn, line 983:
```suggestion
	serviceMonitor := &monitoringv1.ServiceMonitor{ObjectMeta: metav1.ObjectMeta{Name: instance.Name, Namespace: instance.Namespace}}
```

I think using the name of the CR would be a little better for all of these resources, wdyt?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -939,6 +978,50 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+	op, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, serviceMonitor, func(ignore runtime.Object) error {
+		reconcileServiceMonitor(serviceMonitor)
+		return nil
```
grdryn, line 986:
```suggestion
		err := controllerutil.SetControllerReference(instance, serviceMonitor, r.scheme)
		return err
```

^Similar to in the main.go, all of these resources should have owner references set on them

vendor/github.com/integr8ly/grafana-operator/deploy/role.yaml
```diff
@@ -0,0 +1 @@
+roles/role.yaml
```
grdryn, line 1: I'm surprised at this file showing up here, I thought it would only be go files from consumed packages :thinking:

grdryn:
This is good enough that I'm going to merge it now so that I can rebase my changes in #84 on top of it. You can address the comments in a new PR instead of carrying all of this wip along :rocket:

Makefile
laurafitzgerald: Yes it can!

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
laurafitzgerald: Yes, I'm happy to that. The reason for using this change here was the intention that there would be no change to the resources made during this change (to limit upgrade considerations), only that the operator would create it rather than from a file. Seeing as the Name value doesn't get update in the reconcile function I can add some logic in there to remove it if it has the old name.

PR #85, opened 2019-11-15
Make ServiceMonitor a little more robust

grdryn:
JIRA: https://issues.jboss.org/browse/AEROGEAR-10070

For some reason before this change, we were updating the labels on the
Service and relying on the functions from the operator-sdk to re-use
those when creating the ServiceMonitor. Unfortunately, the Service
sometimes didn't get updated, and because this isn't in a controller
that will try indefinitely, the operator needed to be restarted for it
to start again.

This change leaves the Service alone, and just adds the label where we
need it, the ServiceMonitor. I've changed from using the default way
from the operator-sdk scaffold to using the CreateOrUpdate, which
doesn't consider "it already exists" as an error mode, so it will set
it to the expected state if it has drifted.


grdryn:
Covered by #83, closing.

PR #86, opened 2019-11-15
External DB support

jomrazek:
## Motivation
Support for external database connection for UPS
https://issues.jboss.org/browse/AEROGEAR-9950
https://issues.jboss.org/browse/AEROGEAR-9951

## What
Support to choose between connection to external database or creating own 

## Verification Steps
- replace operator image with one build for PR
```sed -i 's/quay\.io\/aerogear\/unifiedpush-operator:0\.3\.0/quay\.io\/aerogear\/unifiedpush-operator:PR-86/g' deploy/operator.yaml```
- update unifiedpush-operator/deploy/push_v1alpha1_unifiedpushserver_cr.yaml
1)externalDB: false
2)externalDB: true (specify database details)
create own database or use this:
```databaseHost: akeatingxwpqhintegreatlycodereadypo-hoaw.cglycytvdjyd.eu-west-1.rds.amazonaws.com
  databaseName: postgres
  databasePassword: 265deb3407cd4b2ebb18e3da7774ca53
  databaseUser: postgres
  databasePort: '5432' 
```
   - make install

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -560,77 +561,90 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
 	}
 	//#endregion
 
 	//#region Postgres Service
-	postgresqlService, err := newPostgresqlService(instance)
-	if err != nil {
-		return reconcile.Result{}, err
+	var postgresqlService *corev1.Service
+	if externaldb {
+		postgresqlService, err = newExternalPostgresqlService(instance)
```
grdryn: Even though we're creating this Service in the case where externaldb is true, I don't believe it's actually being used. Remember that UPS consumes the host from an environment variable that's populated from the secret, and the Secret contains the actual external host in this case, and not this service.

Given that, I'd say we don't need to create this Service when externaldb is true. WDYT?

jomrazek: sounds good, will make change, test it and rebase

laurafitzgerald:
👀

jomrazek:
Change should work ok but I break tests, fixing that.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -609,98 +566,146 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Creating a new Deployment", "Deployment.Namespace", postgresqlDeployment.Namespace, "Deployment.Name", postgresqlDeployment.Name)
+			err = r.client.Create(context.TODO(), postgresqlDeployment)
 			if err != nil {
-				reqLogger.Error(err, "Failed to update Deployment", "Deployment.Namespace", foundPostgresqlDeployment.Namespace, "Deployment.Name", foundPostgresqlDeployment.Name)
 				return reconcile.Result{}, err
 			}
-			return reconcile.Result{Requeue: true}, nil
-		}
-	}
```
grdryn, line 676: The test is failing because this closing brace has been moved down to line 707 (meaning that the creation of the Service is erroneously put into an unrelated condition).

pkg/controller/unifiedpushserver/postgresql.go
```diff
@@ -34,21 +35,35 @@ func newPostgresqlPersistentVolumeClaim(cr *pushv1alpha1.UnifiedPushServer) (*co
 }
 
 func newPostgresqlSecret(cr *pushv1alpha1.UnifiedPushServer) (*corev1.Secret, error) {
-	databasePassword, err := generatePassword()
-	if err != nil {
-		return nil, err
-	}
+	externaldb := &cr.Spec.ExternalDB
+	if !*externaldb {
```
grdryn:
```suggestion
	if !cr.Spec.ExternalDB {
```

- Doesn't need to be a pointer (at least while this is a required field)
- Doesn't need to be an explicit local variable when this is the only place it's called

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -609,98 +566,146 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	// Set UnifiedPushServer instance as the owner and controller
-	if err := controllerutil.SetControllerReference(instance, postgresqlDeployment, r.scheme); err != nil {
-		return reconcile.Result{}, err
-	}
+	externaldb := instance.Spec.ExternalDB
+	if !externaldb {
```
```suggestion
	if !instance.Spec.ExternalDB {
```

deploy/crds/push_v1alpha1_unifiedpushserver_cr.yaml
```diff
@@ -4,3 +4,9 @@ metadata:
   name: example-unifiedpushserver
 spec:
   useMessageBroker: false
+  externalDB: false
+  databaseHost: example-host
```
grdryn: Does it make sense to have these `database*` fields in here when `externalDB` is `false`? Maybe a separate CR example with `externalDB` set to `true` would be better. WDYT?

Also, I wonder if it would be much effort to have the following structure instead :thinking: 

```yaml
  externalDB: false
  database:
    name: example-name
    host: example-host
    port: '5432'
    user: user
    password: password
```

I think it's a little cleaner, wdyt?

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -82,6 +101,8 @@ spec:
               description: UseMessageBroker can be set to true to use managed queues,
                 if you are using enmasse. Defaults to false.
               type: boolean
+          required:
```
grdryn: This should not be required -- it will likely cause upgrades to fail. E.g. if someone has already created a CR where this isn't set (because it doesn't exist), we likely won't be able to "apply" the CRD (or else if we're able to do that, something else will be broken).

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
```diff
@@ -13,6 +13,19 @@ type UnifiedPushServerSpec struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
 
+	//When true, deployement wont be created, service with external db details will be passed to ups
```
grdryn: Notice how in the comment on the other fields, the sentence starts with the name of the thing it's describing. This is a Go convention that we should stick to. See here for more info: https://blog.golang.org/godoc-documenting-go-code

Relevant snippet:
> Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.

pkg/apis/push/v1alpha1/unifiedpushserver_types.go
```diff
@@ -13,6 +13,19 @@ type UnifiedPushServerSpec struct {
...
+	//Database name for external database support
+	DatabaseName string `json:"databaseName,omitempty"`
+	//Database password for external database support
+	DatabasePassword string `json:"databasePassword,omitempty"`
+	//Database user for external database support
+	DatabaseUser string `json:"databaseUser,omitempty"`
+	//Database host for external database support
+	DatabaseHost string `json:"databaseHost,omitempty"`
+	//Database port for external database support
+	DatabasePort string `json:"databasePort,omitempty"`
```
grdryn:
```suggestion
	DatabasePort intstr.IntOrString `json:"databasePort,omitempty"`
```

Allowing either int or string means that users creating a CR don't need to know that they need to quote the string of digits.

grdryn (dismissed):

deploy/crds/push_v1alpha1_unifiedpushserver_crd.yaml
```diff
@@ -69,6 +69,30 @@ spec:
...
+                databaseHost:
+                  description: Database host for external database support
+                  type: string
+                databaseName:
+                  description: Database name for external database support
+                  type: string
+                databasePassword:
+                  description: Database password for external database support
+                  type: string
+                databasePort:
```
jhellar: This value is not used in the secret [1]. Also it is hardcoded in container environment [2].

[1] https://github.com/aerogear/unifiedpush-operator/pull/86/files#diff-a02d8180a2f035b2a960322ad76cd7baR55
[2] https://github.com/aerogear/unifiedpush-operator/blob/8ab2aff005ce988240917ac7e2bd08655f6bd050/pkg/controller/unifiedpushserver/unifiedpushserver.go#L86-L89

jomrazek: good catch :+1:

deploy/crds/push_v1alpha1_unifiedpushserver_cr_with_external_db.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: example-unifiedpushserver
+spec:
+  useMessageBroker: false
+  externalDB: true
+  database: 
+    databaseHost: example-host
+    databaseName: example-name
+    databasePassword: password
+    databaseUser: user
+    databasePort: 5432
```
grdryn: I think it's a bit redundant to have "database" here twice. What do you think about changing it to this instead:

```yaml
  database: 
    host: example-host
    port: 5432
    name: example-name
    user: user
    password: password
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller_test.go
```diff
@@ -47,10 +48,12 @@ func TestReconcileUnifiedPushServer_Reconcile(t *testing.T) {
 	}
 
 	// Check if service has been created
-	service := &corev1.Service{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: pushServerInstance.Name + "-postgresql", Namespace: pushServerInstance.Namespace}, service)
-	if err != nil {
-		t.Fatalf("get service: (%v)", err)
+	if !pushServerInstance.Spec.ExternalDB {
```
grdryn: I think this test should now work without this change, right?

grdryn (approved):
Reviewed and verified! :100: :tada: 

I'll merge this now so that I can rebase my PR on top :+1:

PR #89, opened 2019-12-02
Remove CRDs & make it a namespace operator & add sync period & remove limit on number of UPSes & go back to old UPS version

grdryn:
## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-10091

## What
BURN IT ALL DOWN

## Why
- We don't use those CRDs any longer
- We want to use the older version of UPS
- We don't have any reason to need cluster roles now
- We don't have any reason to limit to a single UPS
- Forcing a sync every 5 minutes sounds like a better idea than waiting for 10 hours or whatever the default is

## How
Deleted some code, updated some other code

## Verification Steps
Add the steps required to check this change. Following an example.
 
1. Deploy the operator
2. Deploy a UPS CR
3. Make sure everything works as expected
4. Delete the UPS CR
5. Make sure everything works as expected

## Checklist:


pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -281,30 +275,6 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
...
 	}
 
-	// look for other unifiedPush resources and don't provision a new one if there is another one with Phase=Complete
-	existingInstances := &pushv1alpha1.UnifiedPushServerList{
-		TypeMeta: metav1.TypeMeta{
-			Kind:       "UnifiedPushServer",
-			APIVersion: "push.aerogear.org/v1alpha1",
-		},
-	}
-	opts := &client.ListOptions{Namespace: instance.Namespace}
```
secondsun, line 291: This breaks like 947 which references opts which hasn't been defined.

secondsun (dismissed):

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: Argh, I obviously tried this out _before_ making this last minute change :roll_eyes: 

Fixed, thanks.

secondsun (approved):
I was able to reproduce your steps and the code looks good.

cmd/manager/main.go
```diff
@@ -109,6 +111,7 @@ func main() {
 		Namespace:          namespace,
 		MapperProvider:     restmapper.NewDynamicRESTMapper,
 		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
+		SyncPeriod:         &syncperiod,
```
davidffrench, line 114: I am not familiar with this, what does it do?

deploy/olm-catalog/unifiedpush-operator/0.4.0/unifiedpush-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,260 @@
...
+        serviceAccountName: unifiedpush-operator
+    strategy: deployment
+  installModes:
+  - supported: true
+    type: OwnNamespace
+  - supported: true
+    type: SingleNamespace
+  - supported: false
+    type: MultiNamespace
+  - supported: true
```
davidffrench, line 243: I believe this also needs to be set to false for AllNamespaces support but I am not too familiar with it.

davidffrench (commented):
The code changes itself look good to me. 2 small comments

deploy/olm-catalog/unifiedpush-operator/0.4.0/unifiedpush-operator.v0.4.0.clusterserviceversion.yaml
grdryn: I don't think so, I think the defaults here are fine, but I'm also not an expert on it. @pb82 I hear you might have thoughts here, does AllNamespaces need to be set to false here?

cmd/manager/main.go
grdryn: I'm pretty sure we've talked about this before, no? :thinking: It's a time period at which the operator will force a reconcile, to make sure everything is correct (in case for some reason an operator missed a trigger, or one wasn't sent or something).

wei-lee (approved):
lgtm

PR #92, opened 2020-01-09
✨ Add optional DatabaseSecret field to CRD

grdryn:
JIRA: https://issues.redhat.com/browse/INTLY-4379

I have deployed this on my cluster for verification, in three different configurations:

- Default, with operator-provisioned postgresql
- Pre-existing external db flow, where the details are provided through the CR
- New external db flow, where the db details are provided through a pre-existing secret

These are deployed to the following namespaces respectively:
- intly-4379a
- intly-4379b
- intly-4379c

See the JIRA issue or contact me for details of the cluster.

Verification steps:
- Make sure that everything is in a ready state & UPS can be accessed.
- Maybe try creating a Push Application
- Make sure there are no persistent errors in the unifiedpush-operator pod (those "unable to decode an Event" log messages caused by enmasse don't count -- we don't use that currently anyway)
- Take a look at the UnifiedPushServer CR, and notice that the `status.secondaryResources` is different (and reasonable) in each scenario


pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -87,6 +88,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(
 		&source.Kind{Type: &pushv1alpha1.UnifiedPushServer{}},
 		&handler.EnqueueRequestForObject{},
+		predicates.ResourceGenerationOrFinalizerChangedPredicate{},
```
grdryn: @davidffrench I've added this predicate here because since I plagiarised the `manageSuccess` function from the new keycloak operator, it seems like reconciliation was thrashing (each `manageSuccess` call naively updates the status block, and that causes the `metadata.resourceVersion` to change, which causes a new reconciliation, which causes a new reconciliation, ad infinitum)

davidffrench: This is correct. The only cluster type I noticed this was an issue on was `minishift`, all other cluster types were clever enough to not update the resourceVersion if there were no changes. 

What are you testing this on?

grdryn: I was testing on minishift. Is it something that was fixed in a particular version of Kubernetes, and that's why you never saw it anywhere else?

davidffrench: I was never able to find out what was different on other clusters compared to minishift. Possibly a parameter on the API server. It was not to do  with a specific kubernetes version from what I saw, the same major.minor version on OSD doesn't have this problem.

philbrookes:
LGTM

philbrookes (approved):

secondsun (approved):
lgtm, I didn't verify the code but trust ger and phil that it was verified ;)

PR #93, opened 2020-03-10
Do not try to watch Enmasse resources if enmasse doesn't support watches

secondsun:
## Motivation
Enmasse 0.30.2 has an incorrect implementation of the watch operation which causes the operator to spam the enmasse service for a list of their resources.  This can cause issues with the enmasse service.  Enmasse 0.31 fixes watches, but hasn't been released yet.

https://issues.redhat.com/browse/AEROGEAR-10176

## What
This PR fixes that issue by detecting an incorrect implementation of enmasse watches.  If an incorrect implementation is detected, then AMQ features will not be enabled in UPS.  Additionally watches for these resources will not be created.

## Why
The failing watch causes the operator to requeue immediately and try again.

## How
There is a method to check if enmasse supports watches correctly.  *The result is cached*.  However, in the future we should be able to remove this once enmasse 0.31+ is more widely adopted

## Verification Steps
 
1. Have a cluster with enmasse 0.30.x running
2. Install the UPS operator
3. Check the operator logs, there should not be errors regarding enmasse
4. Create UnifiedPush with a broker
5. Address, AddressSpace, MessagingUser, etc should *NOT* be created, but UPS should still deploy
6. Upgrade to enmasse 0.31.x (it is in rc right now)
7. Confirm UPS with AMQ deploys successfully

## Checklist:

- [ x ] Code has been tested locally by PR requester


pkg/controller/unifiedpushserver/enmasse.go
```diff
@@ -84,7 +84,7 @@ func newMessagingUser(cr *pushv1alpha1.UnifiedPushServer) (*messaginguserv1beta.
 					Addresses: []string{
 						"*",
 					},
-					Operations: []string{
+					Operations: []messaginguserv1beta.AuthorizationOperation{
```
grdryn, line 87: Is this related to the change in enmasse commit that's being used here? Are there any consequences to this change (I'm not sure if/how it's related to the rest of the PR)?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -73,6 +77,64 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
+// Some versions of enmasse do not handle the address watches correctly
+// this method checks if the watches are handled correctly
+func canWatchEnmasse(cfg *rest.Config) bool {
+	if hasCheckedEnmasse {
+		return canWatchEnmasseVal
+	}
+	dynamicClient, err := dynamic.NewForConfig(cfg)
+	log.Info("Checking if can watch enmasse resources")
+	if err == nil {
+		dynamnicAddressResource := dynamicClient.Resource(schema.GroupVersionResource{
```
```suggestion
		dynamicAddressResource := dynamicClient.Resource(schema.GroupVersionResource{
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -73,6 +77,64 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
 
+// Some versions of enmasse do not handle the address watches correctly
+// this method checks if the watches are handled correctly
+func canWatchEnmasse(cfg *rest.Config) bool {
+	if hasCheckedEnmasse {
+		return canWatchEnmasseVal
+	}
+	dynamicClient, err := dynamic.NewForConfig(cfg)
+	log.Info("Checking if can watch enmasse resources")
+	if err == nil {
```
grdryn, line 88: I think it would be more "Go" if this was `if err != nil {`, since this would keep the happy path aligned to the left rather than nesting it inside an additional block.

See the section titled "Return early rather than nesting deeply" in this recent article: https://dave.cheney.net/2020/02/23/the-zen-of-go#_return_early_rather_than_nesting_deeply

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -73,6 +77,64 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
+				timeout <- true
+			}()
+
+			select {
+			case event := <-watch.ResultChan():
+				if len(string((event).Type)) == 0 {
+					log.Info("Enmasse does not support address watch correctly.")
+					hasCheckedEnmasse = true
+					canWatchEnmasseVal = false
+					return false //The malfunctioning code immedieatly returns an event with an empty type
```
```suggestion
					return false //The malfunctioning code immediately returns an event with an empty type
```

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -73,6 +77,64 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
+		return canWatchEnmasseVal
+	}
+	dynamicClient, err := dynamic.NewForConfig(cfg)
+	log.Info("Checking if can watch enmasse resources")
+	if err == nil {
+		dynamnicAddressResource := dynamicClient.Resource(schema.GroupVersionResource{
+			Group:    "enmasse.io",
+			Version:  "v1beta1",
+			Resource: "addresses",
+		})
```
grdryn, line 93: It looks like the new version of enmasse (1.4.0) uses the same GVK for these resources as the previous version (1.3.1), is that correct? If so, is the difference that they're now instances (CRs) of a CRD (and therefore handled by the k8s api server) rather than pre-CRD resources? If so, I wonder if doing a `client.Get` on the CRD would be an alternative approach here?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -47,8 +49,10 @@ const (
 )
 
 var (
-	cfg = config.New()
-	log = logf.Log.WithName(controllerName)
+	cfg                = config.New()
+	log                = logf.Log.WithName(controllerName)
+	hasCheckedEnmasse  = false
+	canWatchEnmasseVal = false
```
grdryn, line 55: Are these global vars used anywhere apart from inside the `canWatchEnmasse` function?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
```diff
@@ -265,7 +329,7 @@ func (r *ReconcileUnifiedPushServer) Reconcile(request reconcile.Request) (recon
 	}
 
 	//#region AMQ resource reconcile
-	if instance.Spec.UseMessageBroker {
+	if instance.Spec.UseMessageBroker && canWatchEnmasse(r.config) {
```
grdryn, line 331: Is it right to just ignore the spec when we can't create a watch? We can still create and use the resources, even if we're not going to watch them, right? I haven't fully thought this through, so I might well be wrong :)

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: not really.  The dynamic client is the only way to get access to the event channel and it is the immediate useless close of the event channel that signals the error has happened.

pkg/controller/unifiedpushserver/enmasse.go
secondsun: Enmasse change they type signature; it is still a string underneath.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: They are not.  It would probably be more golike to wrap the canWatchEnmasse in its own pacake to isolate those variables.

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
secondsun: Well creating the child address causes the operator to start watching those resources.  I'm not really sure where, but the errors star appearing and the operator goes into an immediate infinite reconcile loop.  Since that behavior will bog down a enmasse server pre 0.31 I thought it was best to just disable it.

pkg/controller/unifiedpushserver/enmasse.go
grdryn: Probably a more important question then: why is the version of the enmasse dependency changing?

pkg/controller/unifiedpushserver/unifiedpushserver_controller.go
grdryn: What if they were just declared at the top of the function? Or does the pre-existing state matter (i.e. what it may have set them to last time it was called)?

secondsun: The idea is that we only need to make that check once since it needs a brief timeout window.  That timeout, however, is blocking so these variable make it check only once
 at startup.

pkg/controller/unifiedpushserver/enmasse.go
secondsun: Because updates?

secondsun:
We're not going with the approach any more.  See the JIRAs

PR #96, opened 2020-04-03
Updating Enmasse to work with AMQ Online 1.4

secondsun:
## Motivation
AMQ Online has been updated to 1.4 which fixes a lot of issues we've had.  This PR brings those updates to unifiedpush-operator.
## What
 * enmasse was updated to the latest version in master 
 * type related code breaks were fixed, but they didn't change logic.

## Why
There were many issues with AMQ Online.  The ones that affected us were
 * infinite logs with bad watches (this is fixed in 1.4)
 * addresses hanging on creation (this should have been fixed in 1.4 as well)

## Verification Steps
Add the steps required to check this change. Following an example.
 
1. Install this operator on a cluster with AMQ Online 1.4
2. Create a ups resource with useMessageBroke : true
3. Ensure that the ups, messaging user, addresspace, and addresses all are created and running
4. Ensure that UPS works as normal
## Checklist:

- [ x ] Code has been tested locally by PR requester

## Progress

- [x] Finished task

## Additional Notes

I can set up a cluster if necessary; the code was run locally.


Gopkg.lock
```diff
@@ -137,7 +137,7 @@
 
 [[projects]]
   branch = "master"
```
grdryn: I wonder if there's a release tag/branch that we could use instead (probably would be specified in the Gopkg.toml)

grdryn (commented):
LGTM. I'll try to verify it now on an OpenShift 4 cluster, and then will give the green tick

If you're planning to do a release of this (which we'll probably need for integreatly?), don't forget to generate the CSV as part of that process :+1:

grdryn:
I ran this in an OpenShift 4 cluster that has integreatly installed (which includes enmasse 1.4), and it (enmasse) consistently failed to get all the queue addresses to a ready state. What's stranger is that it was always the same two queues that weren't getting ready:

- ups.apnspushmessagequeue
- ups.apnstokenbatchqueue

I think on the first occasion I deleted them, causing the unifiedpush operator to recreate them, and they got to the ready state then; but that didn't work on subsequent attempts.

Any ideas why that might be?

grdryn (approved):
I've re-verified this after you fixed the Gopkg.toml/Gopkg.lock

Given that the issue with the queues getting stuck is out of our control, and existed previously; I think this change should be merged and released. :+1:

secondsun:
Landed!

PR #98, opened 2020-06-19
fix: adding the v in version

secondsun:
## Motivation
See INTLY-8590

## What
The version field needs to match the version tag for purposes of linking. 

## Why
The wrong version tag was causing incorrect links to be generated

## Verification Steps
This will have to be tested when a new release is made.

1. Deploy the operator
2. Make sure the links in the Prometheus monitor work.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "0.5.0"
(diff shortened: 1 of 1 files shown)


grdryn (approved):
:shipit: 

Edit: I haven't tried this out.

PR #102, opened 2020-07-28
Update constants.go

secondsun:
see INTLY-7485


jhellar:
Tests are not passing because Jenkinsfile in 0.4.x branch does not contain `QUAY_TOKEN_ID` [1]. It is only defined in master [2]. That should not be a problem though, as it is only used to cleanup the `candidate` image from quay.io.

[1] https://github.com/aerogear/unifiedpush-operator/blob/0.4.x/Jenkinsfile
[2] https://github.com/aerogear/unifiedpush-operator/blob/master/Jenkinsfile#L26

jhellar:
Verified backup in RHMI.

